PR #23, opened 2015-07-20
sql: parse Mysql INTERVALs

cyberhuman:
Basic support for Mysql's INTERVAL keyword.
Any clues what to do about "year", INTERVAL_UNIT conflicting with "year", T_DATETIME?

### DIFF
--- a/lib/sql_lexer.mll
+++ b/lib/sql_lexer.mll
@@ -121,6 +121,27 @@ let keywords =
    "action", ACTION;
    "no", NO;
    "is", IS;
+   "interval", INTERVAL;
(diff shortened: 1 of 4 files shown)


ygrek:
as sqlgg is moving towards proper sql parser with AST - will need to distinguish INTERVAL_UNIT's anyway - should add separate token for each keyword and then can reuse year token

cyberhuman:
Updated accordingly.

ygrek:
d1fa845dd36ff279285e7429fcddc5960dbdf44b
Thanks

PR #60, opened 2018-12-19
Monadic io

cyberhuman:
Turned off by default; use `-io` to turn on. And unfortunately compatible traits are not implemented yet.
Extra: `-no-header-timestamp` not to put timestamp in a generated file header.


ygrek:
thanks!

PR #61, opened 2018-12-27
monadic io traits for ocaml-mariadb

cyberhuman:


impl/ocaml/mariadb/sqlgg_mariadb.ml
```diff
@@ -0,0 +1,231 @@
...
+  type field
+  type value
+  val of_field : field -> t
+  val to_value : t -> value
+end
+
+module type Types = sig
+  type field
+  type value
+  module Bool : Value with type field = field and type value = value
```
ygrek:
```suggestion
  module type Value = Value with type field = field and type value = value
  module Bool : Value
  module Int : Value
  module Float : Value
  module Text : Value
  module Datetime : Value
  module Any : Value
```

impl/ocaml/mariadb/sqlgg_mariadb.ml
```diff
@@ -0,0 +1,231 @@
...
+  M.Res.fetch row_array >>=
+  check >>=
+  function
+  | Some row -> return (Some (callback row))
+  | None -> return None
+
+end
+
+module Default(IO : Sqlgg_io.M)(M : Mariadb.Nonblocking.S with type 'a future = 'a IO.future) =
+  Make(IO)(M)
```
ygrek: this probably intended to apply `Default_types`?

PR #103, opened 2021-06-02
Update traits to track read/write nature of ops

jchavarri:
Adds a type variable to `connection` that is used later on to track the nature of the queries: either readonly, or readwrite.

More work is prob need to either:
- make sure it's released as major version
- duplicate traits module and introduce a flag to choose from which traits the exported code will read from


impl/ocaml/sqlgg_traits.ml
```diff
@@ -79,25 +79,25 @@ module type M = sig
     Perform query (cardinality "any") and return results via callback for each row
     @raise Oops on error
   *)
-  val select : connection -> string -> (statement -> result) -> (row -> unit) -> unit
+  val select : [>`RO] connection -> string -> (statement -> result) -> (row -> unit) -> unit
```
ygrek, line 82: I think it will be better to have two separate abstract types in Traits - `ro_connection` and `rw_connection` and let specific implementations decide if they want subtyping relationship, will that work?

jchavarri: I took a stab at this but it's not trivial. If abstract types have no param then one has to assign like:
```
  type nonrec 'a connection = 'a connection
  type ro_connection = [`RO] connection
```

using closed variants. This means that values that would match using open variants now fail, e.g.:

```
Error: This expression has type rw_connection
       but an expression was expected of type [> `RO ] connection
```

and removes the possibility of leveraging inference to derive types like ``[> `RO | `WR ]`` when reads and writes are inside same function.

A couple of ideas:

1. Add a second copy of `sqlgg_traits` that has everything needed from type perspective to use ro/rw connections, allow to choose this alternate traits module with flag `-readwrite` for example.
2. If the ro/rw use case is very specific, leave things as they were before but add a flag `-custom_traits` that allows to pass the name of the custom traits module that will be used in application, moving this problem (and any others) to userland altogether
3. Explore using abstract types with params, e.g. `'a ro_connection` and `'a rw_connection`.

jchavarri: I took a stab at 3, but it seems it would be a net worse solution:

- Users that don't need readonly/readwrite tracking now have to unify two types, instead of one before.
- Users that need readonly/readwrite tracking have to redefine the types to add the constraints.

Options 1 and 2 are also more cumbersome, as they leak to configuration.

@ygrek do you have any alternatives? Or otherwise, would it be ok to go with the current proposal? Users that don't need ro/rw can easily override `connection` type, e.g.:

https://github.com/ygrek/sqlgg/blob/a05d173708c1c32ab34a95ff60cabba5aa9a513d/impl/ocaml/mysql/sqlgg_mysql.ml#L67

Users that need readonly/readwrite can use the library as is, and benefit from the upsides, as it allows type checker to infer types correctly with very few / no type annotations.

ygrek: > If the ro/rw use case is very specific, leave things as they were before but add a flag -custom-traits that allows to pass the name of the custom traits module that will be used in application, moving this problem (and any others) to userland altogether. example:
> sqlgg -gen caml -custom-traits ReadWriteTraits -params named demo.sql

Actually it is already fully in userland, user can chose to not link with sqlgg runtime library but provide sqlgg_traits module and specific implementations as they see fit. But indeed should ship such features natively, especially they don't pose much burden on new implementations.

PR #104, opened 2021-11-09
Support for arguments in IN expressions

mfp:
Perform textual substitution to generate the SQL statement corresponding to the arguments passed to (NOT) IN expressions.

Given

```sql
-- @create_table
CREATE TABLE foo(
    id INTEGER PRIMARY KEY,
    foo TEXT NULL
);
-- @find
SELECT * from foo
WHERE id IN @ids;
```
generate code like

```ocaml
    let find db ~ids callback =
      let __sqlgg_sql =
        let replace_all ~str ~sub ~by =
          let rec loop str = match ExtString.String.replace ~str ~sub ~by with
          | true, str -> loop str
          | false, s -> s
          in loop str
        in
        let sql = ("SELECT * from foo\n\
WHERE id IN @@_sqlgg_ids@@") in
        let sql = replace_all ~str:sql ~sub:("@@_sqlgg_ids@@") ~by:( "(" ^ String.concat ", " (List.map T.Types.Int.to_literal ids) ^ ")") in
        sql
      in
      let invoke_callback stmt =
        callback
          ~id:(T.get_column_Int stmt 0)
          ~foo:(T.get_column_Text_nullable stmt 1)
      in
      let r_acc = ref [] in
      IO.(>>=) (T.select db __sqlgg_sql T.no_params (fun x -> r_acc := invoke_callback x :: !r_acc))
      (fun () -> IO.return (List.rev !r_acc))
```

### Limitations
- the code generated for empty lists is syntactically invalid for MySQL and MariaDB. In order to solve this, the whole expression would need to be replaced by a suitable `TRUE` (for `NOT IN`) or `FALSE` (for `IN`), but this requires extra work in the lexer or new syntax to determine the initial position (the end is the parameter's)
- not supported in cxx and java backends

### DIFF
--- a/impl/ocaml/sqlgg_runtime.ml
+++ b/impl/ocaml/sqlgg_runtime.ml
@@ -1,0 +1,10 @@
+
+let replace_all_chars ~str ~sub ~by =
+  let rec get_chunks s sub acc off =
+    if off >= String.length s then List.rev ("" :: acc)
+    else
+      match String.index_from_opt s off sub with
+      | None -> List.rev (String.sub s off (String.length s - off) :: acc)
+      | Some off' -> get_chunks s sub (String.sub s off (off' - off) :: acc) (off' + 1)
+  in
+  String.concat by (get_chunks str sub [] 0)
--- a/impl/ocaml/sqlgg_traits.ml
+++ b/impl/ocaml/sqlgg_traits.ml
@@ -13,7 +13,7 @@
   For more information, please refer to <http://unlicense.org/>
 *)
 
-module type Value = sig type t end
+module type Value = sig type t val to_literal : t -> string end
 
 module type M = sig
 
--- a/impl/ocaml/sqlgg_traits.ml
+++ b/impl/ocaml/sqlgg_traits.ml
@@ -13,7 +13,12 @@
   For more information, please refer to <http://unlicense.org/>
 *)
 
-module type Value = sig type t val to_literal : t -> string end
+module type Value = sig
+  type t
+
+  (** Return the literal representation of the value in the SQL dialect. *)
+  val to_literal : t -> string
+end
 
 module type M = sig
 
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -362,7 +362,7 @@ expr:
     | e1=expr mnot(IN) l=sequence(expr) { poly Bool (e1::l) }
     | e1=expr mnot(IN) LPAREN select=select_stmt RPAREN { poly Bool [e1; Select (select, `AsValue)] }
     | e1=expr IN table=table_name { Tables.check table; e1 }
-    | e1=expr mnot(IN) p=PARAM { Fun (InF (Var 0, Var 1), [ e1; Inparam (new_param p Any) ]) }
+    | e1=expr mnot(IN) p=PARAM { poly Bool [ e1; Inparam (new_param p Any) ] }
     | LPAREN select=select_stmt RPAREN { Select (select, `AsValue) }
     | p=PARAM { Param (new_param p Any) }
     | p=PARAM parser_state_ident LCURLY l=choices c2=RCURLY { let { label; pos=(p1,_p2) } = p in Choices ({ label; pos = (p1,c2+1)},l) }
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -362,6 +362,7 @@ expr:
     | e1=expr mnot(IN) l=sequence(expr) { poly Bool (e1::l) }
     | e1=expr mnot(IN) LPAREN select=select_stmt RPAREN { poly Bool [e1; Select (select, `AsValue)] }
     | e1=expr IN table=table_name { Tables.check table; e1 }
+    | e1=expr mnot(IN) p=PARAM { Fun (InF (Var 0, Var 1), [ e1; Inparam (new_param p Any) ]) }
     | LPAREN select=select_stmt RPAREN { Select (select, `AsValue) }
     | p=PARAM { Param (new_param p Any) }
     | p=PARAM parser_state_ident LCURLY l=choices c2=RCURLY { let { label; pos=(p1,_p2) } = p in Choices ({ label; pos = (p1,c2+1)},l) }
--- a/lib/syntax.ml
+++ b/lib/syntax.ml
@@ -31,6 +31,7 @@ let rec get_params_q (e:expr_q) =
   let rec loop acc e =
     match e with
     | `Param p -> Single p::acc
+    | `Inparam p -> SingleIn p::acc
     | `Func (_,l) -> List.fold_left loop acc l
     | `Value _ -> acc
     | `Choice (p,l) -> Choice (p, List.map (fun (n,e) -> Simple (n, Option.map get_params_q e)) l) :: acc
@@ -47,6 +48,7 @@ let rec is_grouping = function
 | Param _
 | Column _
 | Select _
+| Inparam _
 | Inserted _ -> false
 | Choices (p,l) ->
   begin match list_same @@ List.map (fun (_,expr) -> Option.map_default is_grouping false expr) l with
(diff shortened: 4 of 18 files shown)


lib/syntax.ml
```diff
@@ -177,6 +187,32 @@ and assign_types expr =
             convert ret, List.map convert args
           else
             fail "types do not match : %s" (show ())
+        | InF (lhs, rhs), _ ->
```
ygrek, line 190: this whole branch seems equivalent to above branch applied to `F (Bool, [lhs; Tuple rhs])`
ie either it can be factored better or `InF` can be completely replaced with `F`

src/gen.ml
```diff
@@ -86,6 +86,18 @@ let substitute_vars s vars subst_param =
           parami + 1
       in
       loop acc i2 parami tl
+    | SingleIn param :: tl ->
```
ygrek, line 89: it seems SingleIn semantics more precisely is Single param with text substitution (as opposed prepared statements parameter for Single), so lets rename to say `SingleText`?

src/gen_caml.ml
```diff
@@ -234,16 +235,23 @@ let output_params_binder _ vars =
...
+  List.iteri (fun index v -> set_var index v) vars;
   output "T.finish_params p";
   dec_indent ();
   output "in";
   "set_params"
 
+let exclude_in_vars l =
+  List.filter
+    (function
+      | Single _ | Choice _ -> true
```
ygrek, line 247: this is interesting, observations:
1. choice should be recursed
2. choice itself is also text-time substitution, so by itself it should lead to T.no_params but in majority of cases the branches will have internal parameters
3. i wonder if it is only an issue of pretty output but not correctness because output_params_binder with empty list as input will just  generate identity function

src/gen_caml.ml
```diff
@@ -268,10 +276,32 @@ let make_sql l =
...
 
+let in_var_module label = function
+  | Sql.Type.Tuple _ | Any | Unit _ as typ -> failwith @@ sprintf "invalid IN var type %s in @%s" (Sql.Type.show typ) label
+  | typ -> Sql.Type.to_string typ
+
+let gen_in_substitution var =
+  if Option.is_none var.id.label then failwith "empty label in IN param";
+  match var.typ with
+  | Sql.Type.Unit _ | Int | Text | Blob | Float | Bool | Datetime | Decimal | Any | Tuple (Any | Tuple _) as typ ->
+    failwith @@ sprintf "invalid inferred IN var type %s (%s)" (Sql.Type.show typ) (Option.get var.id.label)
```
ygrek, line 287: this cannot possibly happen, by construction, ie it is more of assert than error for the user? (nothing to change, just asking to understand)

src/gen_caml.ml
```diff
@@ -290,8 +320,13 @@ let generate_stmt style index stmt =
     output "    in loop str";
     output "  in";
     output "  let sql = %s in" sql;
-    List.iter begin fun var ->
-      output "  let sql = replace_all ~str:sql ~sub:(\"%%%%%s%%%%\") ~by:%s in" var var;
+    List.iter begin function
+      | `Var var ->
+        output "  let sql = replace_all ~str:sql ~sub:(\"%%%%%s%%%%\") ~by:%s in" var var;
```
ygrek, line 325: we should use same pattern and unify these two cases one day (but that day is not today)

lib/sql_parser.mly
```diff
@@ -362,6 +362,7 @@ expr:
     | e1=expr mnot(IN) l=sequence(expr) { poly Bool (e1::l) }
     | e1=expr mnot(IN) LPAREN select=select_stmt RPAREN { poly Bool [e1; Select (select, `AsValue)] }
     | e1=expr IN table=table_name { Tables.check table; e1 }
+    | e1=expr mnot(IN) p=PARAM { Fun (InF (Var 0, Var 1), [ e1; Inparam (new_param p Any) ]) }
```
ygrek, line 365: core of my feedback to this PR is if this can be replaced by smth like `Fun (F (Bool, [Var 0, Var 1]), [e1; Param (new_param p (Tuple Any)])` and I suspect not immediately possible

ygrek (commented):
It seems to me it is (almost?) possible to express typing with existing `F` but I am not 100% sure and if it is not possible I am ok to merge it in the current form and improve typing later.

impl/ocaml/sqlgg_traits.ml
```diff
@@ -13,7 +13,7 @@
   For more information, please refer to <http://unlicense.org/>
 *)
 
-module type Value = sig type t end
+module type Value = sig type t val to_literal : t -> string end
```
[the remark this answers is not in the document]
mfp: yes, will comment

> should it take sql variant used or this is unlikely to be relevant?

I'm confused: there's an implementation of the traits for each supported DB (`Sqlgg_mariadb`, `Sqlgg_mysql`, `Sqlgg_sqlite3`), so each provides the appropriate `to_literal` functions for its own dialect?
(gotta double check the implementations btw.)

lib/syntax.ml
mfp: my first attempt used `F (Bool, _)` but I couldn't get type unification to work, will try again

impl/ocaml/sqlgg_traits.ml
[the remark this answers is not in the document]
ygrek: indeed, you are right

lib/sql_parser.mly
mfp: That was my first attempt (which failed), but maybe I can extend the type variable language to be able to represent the polymorphic type `Tuple (Var 0)`, thus getting rid `InF` case. The direct encoding would require using a polymorphic variant with open recursion.

src/gen_caml.ml
mfp: Yes, basically "in variables" must have `Tuple x` type (where x is not a Tuple itself), so this is just an assert with a more informative message for the `sqlgg` developer.
 
This could be removed by making `param` polymorphic and switching to polymorphic variants in the type language, but it requires extra annotations (or patience) to type.

Another idea is adding a "is IN var" flag to `param`. That might allow to reuse `F` as is.

mfp:
I managed to get rid of `InF` by removing the `Tuple _` type, and letting the types unify, so we treat `x IN @foo` as `'a -> 'a -> bool` and only at the very end, when the code is emitted, we separate regular variables from those carrying sets.

impl/ocaml/sqlgg_runtime.ml
```diff
@@ -0,0 +1,10 @@
+
+let replace_all_chars ~str ~sub ~by =
+  let rec get_chunks s sub acc off =
+    if off >= String.length s then List.rev ("" :: acc)
+    else
+      match String.index_from_opt s off sub with
+      | None -> List.rev (String.sub s off (String.length s - off) :: acc)
+      | Some off' -> get_chunks s sub (String.sub s off (off' - off) :: acc) (off' + 1)
+  in
```
ygrek, line 9: this reimplements String.split_on_char which is in stdlib, why

src/gen_xml.ml
```diff
@@ -62,6 +62,11 @@ let start () = ref [], ref []
 let generate_code (x,_) index stmt =
   let name = choose_name stmt.props stmt.kind index in
   let input = Node ("in",[],params_to_values @@ params_only stmt.vars) in
+  let input_sets =
+    match inparams_only stmt.vars with
+    | [] -> None
+    | _ -> Some (Node ("in-set",[],params_to_values @@ inparams_only stmt.vars))
```
ygrek, line 68: I think I like `<in value= type= set=true>` more (where `set` is optional when false)

src/gen_caml.ml
```diff
@@ -240,10 +241,23 @@ let output_params_binder _ vars =
   output "in";
   "set_params"
 
+let rec exclude_in_vars l =
```
ygrek, line 244:
```suggestion
let rec exclude_text_vars l =
```

src/gen_caml.ml
```diff
@@ -268,10 +282,38 @@ let make_sql l =
...
+  if Option.is_none var.id.label then failwith "empty label in IN param";
+  sprintf {code| "(" ^ String.concat ", " (List.map T.Types.%s.to_literal %s) ^ ")"|code}
+    (in_var_module (Option.get var.id.label) var.typ)
+    (Option.get var.id.label)
+
+let extract_in_subst vars =
+  let rec extract acc = function
+    | [] -> List.rev acc
+    | Single _ :: tl -> extract acc tl
+    | SingleText p :: tl -> extract (`InVar p :: acc) tl
```
ygrek, line 297: this is backwards now, SingleText makes an impression of generic text substitution but generates specifically only `IN`

src/gen.ml
```diff
@@ -86,6 +86,18 @@ let substitute_vars s vars subst_param =
           parami + 1
       in
       loop acc i2 parami tl
+    | SingleText param :: tl ->
+      let (i1,i2) = param.id.pos in
+      let acc, parami =
+        match param.id.label with
+        | None -> failwith "empty label in IN param"
+        | Some label ->
+          Static (sprintf "@@_sqlgg_%s@@" label) ::
```
ygrek, line 95: actually proper way to substitute would be to extend this Static/Dynamic machinery (have one more Dynamic flavor)

ygrek:
Glorious, thanks!

impl/ocaml/sqlgg_runtime.ml
mfp: The real reason is that I forgot it. The post hoc one is that `split_on_char` is 4.03+. Is 4.02 still supported?

ygrek: I am happy to drop 4.02 in favor of split_on_char :muscle:

PR #106, opened 2021-11-09
sql: fix ORDER BY with multiple params

mfp:
ditto


ygrek:
the fix is the order of parameters in generated code?

mfp:
no, it was emitting broken code that didn't compile

ygrek:
Thanks!
Cherry-picked in 647813497bd9d3c8255679cd475b68bca273e2d1

PR #107, opened 2021-11-22
impl: decrease allocation in string to_literal

mfp:
- faster + keeps 4.02 compat
- breaks the API but it was barely made public? (add alias to `String.split_on_char` and abandon 4.02 if it matters)


impl/ocaml/sqlgg_runtime.ml
```diff
@@ -1,10 +1,32 @@
...
+  Buffer.contents b
+
+(* cf. https://sqlite.org/lang_expr.html "Literal Values"
+    "A string constant is formed by enclosing the string in single quotes
+    ('). A single quote within the string can be encoded by putting two
+    single quotes in a row - as in Pascal. C-style escapes using the
+    backslash character are not supported because they are not standard
+    SQL." *)
+let sqlite3_string_to_literal s =
+  let b = Buffer.create (String.length s + String.length s lsl 2) in
```
ygrek, line 24: > lsl 2

lets not go overboard, `* 4` is easier to read. But why such a huge buffer? it will never get fully used

ygrek (changes requested):
breaking api doesn't matter, it was not part of release yet
the only qn is if these funcs should just be in traits? duplicated for mysql/mariadb but that's fine?

impl/ocaml/sqlgg_runtime.ml
mfp: that was me trolling for the laughs with the bit shifting and shooting myself in the foot; changed to `len + len / 4`

mfp:
Will move the functions tomorrow.

ygrek (approved):

PR #108, opened 2021-11-24
Support for MySQL multi-table DELETE FROM

mfp:
refer to https://dev.mysql.com/doc/refman/5.7/en/delete.html

Syntax:
```sql
DELETE t, foo FROM tbl1 t JOIN foo ON t.foo_id = foo.id JOIN bar ON t.bar_id = bar.id
WHERE ...; 
```


lib/sql.ml
```diff
@@ -330,6 +330,7 @@ type stmt =
 | CreateIndex of string * table_name * string list (* index name, table name, columns *)
 | Insert of insert_action
 | Delete of table_name * expr option
+| DeleteMulti of table_name * table_name list * nested * expr option
```
ygrek, line 333: this table_name is not special, I understand it is ad hoc typed promise of not empty list, I would rather see one list

lib/stmt.ml
```diff
@@ -17,7 +17,7 @@ type kind = | Select of cardinality
             | Create of Sql.table_name
             | CreateIndex of string
             | Update of Sql.table_name option (** name for single-table UPDATEs *)
-            | Delete of Sql.table_name
+            | Delete of Sql.table_name option (** name for single-table DELETEs *)
```
ygrek, line 20: lets be more precise and have table_name list here

src/gen_xml.ml
```diff
@@ -83,7 +83,8 @@ let generate_code (x,_) index stmt =
     | CreateIndex t    -> ["kind", "create_index"; "target",t;"cardinality","0"]
     | Update None      -> ["kind", "update"; "cardinality", "0"]
     | Update (Some t)  -> ["kind", "update"; "target", Sql.show_table_name t; "cardinality", "0"]
-    | Delete t         -> ["kind", "delete"; "target", Sql.show_table_name t; "cardinality", "0"]
+    | Delete None      -> ["kind", "delete"; "cardinality", "0"]
+    | Delete (Some t)  -> ["kind", "delete"; "target", Sql.show_table_name t; "cardinality", "0"]
     | Alter t          -> ["kind", "alter"; "target", String.concat "," @@ List.map Sql.show_table_name t; "cardinality", "0"]
```
ygrek, line 88: and handle multi-table delete same as multi-table alter

ygrek (changes requested):

ygrek (approved):

PR #109, opened 2022-01-07
Support empty sets in (x IN @foo) exprs

mfp:
- parentheses around the whole expression are now required
- replaces the whole expression with `TRUE` (for `NOT IN`) or `FALSE` when the list is empty -- note that this is not entirely correct, as the left operand could be NULL
- the LHS operand can have variables, be a choice, etc. -- however, it must be possible to infer a concrete type for the elements: `(COALESCE(bar, @baz) IN @foo)` will work, but `(@foo in @foos)` won't

This is what the generated code looks like:

```sql
--@find_var2
SELECT * FROM foo WHERE (COALESCE(bar, CONCAT(@baz, @x)) IN @ids);
```

```ocaml
    let find_var2 db ~baz ~x ~ids callback =
      let invoke_callback stmt =
        callback
          ~id:(T.get_column_Int stmt 0)
          ~foo:(T.get_column_Text_nullable stmt 1)
          ~bar:(T.get_column_Text_nullable stmt 2)
      in
      let set_params stmt =
        let p = T.start_params stmt (0 + (match ids with [] -> 0 | _ :: _ -> 2)) in
        begin match ids with 
        | [] -> ()
        | _ :: _ ->
          T.set_param_Text p baz;
          T.set_param_Text p x;
          ()
        end;
        T.finish_params p
      in
      let r_acc = ref [] in
      IO.(>>=) (T.select db ("SELECT * FROM foo WHERE " ^ (match ids with [] -> "FALSE" | _ :: _ -> "(COALESCE(bar, CONCAT(" ^ "?" ^ ", " ^ "?" ^ ")) IN " ^  "(" ^ String.concat ", " (List.map T.Types.Text.to_literal ids) ^ ")" ^ ")")) set_params (fun x -> r_acc := invoke_callback x :: !r_acc))
      (fun () -> IO.return (List.rev !r_acc))

```


mfp:
idk why the build fails in AppVeyor
```
# File "lib/sql_parser.mly", line 377, characters 48-66:
# Error: This expression has type 'a * 'b
#        but an expression was expected of type Sqlgg.Sql.select_full
```
The (unmodified) offending line is
```
| LPAREN select=select_stmt RPAREN { Select (select, `AsValue) } 
```
and the `select_stmt` production returns `{ select = ($1, other); order=o; limit=lim; }` which is a `Sql.select_full`. :thinking:

mfp:
earlier AppVeyor build fail fixed by merge against upstream master

lib/sql.ml
```diff
@@ -242,6 +242,7 @@ type ctor =
 and var =
 | Single of param
 | SingleIn of param
+| ChoiceIn of param_id * bool * var list
```
rr0gi, line 245:
```suggestion
| ChoiceIn of { param : param_id; default : bool; vars : var list }
```

lib/sql_parser.mly
```diff
@@ -369,7 +369,11 @@ expr:
     | e1=expr mnot(IN) l=sequence(expr) { poly Bool (e1::l) }
     | e1=expr mnot(IN) LPAREN select=select_stmt RPAREN { poly Bool [e1; SelectExpr (select, `AsValue)] }
     | e1=expr IN table=table_name { Tables.check table; e1 }
-    | e1=expr mnot(IN) p=PARAM { poly Bool [ e1; Inparam (new_param p Any) ] }
+    | lp=LPAREN e1=expr b=in_or_not_in p=PARAM rp=RPAREN
```
rr0gi, line 372: so the parentheses requirement comes only from the need to infer locations? we can use http://gallium.inria.fr/~fpottier/menhir/manual.html#sec52 then

lib/sql.ml
rr0gi: I wonder f ChoiceIn can be somehow avoided with new variant(s) of `ctor` - ie one representing "ensure parameter is non empty list" case. Not sure if it is worth it (supposedly it will simplify handling of params but will require more attention at time of destructing user value - this ctor cannot be mixed with other ctors)

rr0gi (changes requested):

lib/sql_parser.mly
mfp: figured it out, required some care to preserve the lexeme start in the lexer

rr0gi (approved):

ygrek (approved):

PR #110, opened 2022-05-09
Batch insert

mfp:
Support batch insert (using a tuple list) with syntax `INSERT INTO foo(x, y) VALUES @foos` or `INSERT INTO foo VALUES @foos`.

Notes
- empty tuple lists are supported
- the generated code performs textual substitution and builds the SQL in linear space/time
- the param is shown as having type `Any` in the XML output

The generated code looks like

```ocaml
  let insert_set db ~foos ~duplicate =
    let set_params stmt =
      begin match duplicate with None -> T.set_param_null p | Some v -> T.set_param_Text p v end;
      T.finish_params p
    in
    ( match foos with [] -> T.execute db "-- empty insert" T.no_params | _ :: _ -> T.execute db ("INSERT INTO foo(id, foo) VALUES " ^ (let _sqlgg_b = Buffer.create 13 in List.iteri (fun _sqlgg_idx (id, foo) -> Buffer.add_string _sqlgg_b (if _sqlgg_idx = 0 then "(" else ", ("); Buffer.add_string _sqlgg_b (T.Types.Int.to_literal id); Buffer.add_string _sqlgg_b ", "; Buffer.add_string _sqlgg_b (T.Types.Text.to_literal foo); Buffer.add_char _sqlgg_b ')') foos; Buffer.contents _sqlgg_b) ^ "\n\
ON DUPLICATE KEY UPDATE foo = ?") set_params )
```

given

```sql
CREATE TABLE foo(
    id INTEGER PRIMARY KEY,
    foo TEXT NULL
);

-- @insert_set
INSERT INTO foo(id, foo) VALUES @foos
ON DUPLICATE KEY UPDATE foo = @duplicate;
```

### DIFF
--- a/src/gen_xml.ml
+++ b/src/gen_xml.ml
@@ -47,24 +47,30 @@ let _ =
 let comment (x,_) fmt = Printf.ksprintf (fun s -> x := Comment s :: !x) fmt
 let empty_line _ = ()
 
-let value ?(inparam=false) ?(tuplelist=false) v =
+let value ?(inparam=false) v =
   let attrs =
     List.concat
       [
-        ["name",v.vname; "type",v.vtyp];
-        if inparam then ["set","true"] else [];
-        if tuplelist then ["tuplelist","true"] else [];
+        ["name",v.vname; "type", if inparam then sprintf "set(%s)" v.vtyp else v.vtyp];
         if v.nullable then ["nullable","true"] else [];
       ]
   in
   Node ("value", attrs, [])
+
+let tuplelist_value_of_param = function
+  | Sql.Single _ | SingleIn _ | Choice _ | ChoiceIn _ -> None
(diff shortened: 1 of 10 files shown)


src/gen_xml.ml
```diff
@@ -64,14 +81,18 @@ let get_sql_string stmt =
   let rec map i = function
   | Static s -> s
   | SubstIn param -> "@@" ^ show_param_name param i (* TODO join text and prepared params earlier for single indexing *)
+  | SubstTuple (id, _) -> "@@@" ^ make_param_name i id
```
ygrek, line 84: :doge:

ygrek (approved):
some opportunities for factoring start to appear in printer, but it is separate concern from this PR
thanks!

PR #115, opened 2023-03-10
Fix DATE signature and generated code for multi-value inserts

mfp:
.

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -424,7 +424,8 @@ let () =
   ["least";"greatest";"coalesce"] ||> multi_polymorphic;
   "strftime" |> exclude 1; (* requires at least 2 arguments *)
   ["concat";"strftime"] ||> multi ~ret:(Typ Text) (Typ Text);
-  ["date";"time"] ||> monomorphic Text [Datetime];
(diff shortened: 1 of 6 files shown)


lib/sql.ml
```diff
@@ -424,7 +424,7 @@ let () =
   ["least";"greatest";"coalesce"] ||> multi_polymorphic;
   "strftime" |> exclude 1; (* requires at least 2 arguments *)
   ["concat";"strftime"] ||> multi ~ret:(Typ Text) (Typ Text);
-  ["date";"time"] ||> monomorphic Text [Datetime];
+  ["date";"time"] ||> monomorphic Datetime [Datetime];
```
rr0gi, line 427: is it correct for `time`?

rr0gi (changes requested):
need to update tests (`make test` and commit)

lib/sql.ml
mfp: I'm not sure (didn't see any actual use of TIME). I changed this to leave it as it was.

also
> Extracts the time part of the time or datetime expression expr and returns it as a string.

PR #118, opened 2023-10-27
Reuse inferred param_types in the usage below the body

jongleb:
We let users work with a variable as if it has multiple types.

For example:

**Our schema:** 

```sql
-- @create_some_table
CREATE TABLE IF NOT EXISTS `some_table` (
 `id` int NOT NULL,
 `some_field` TEXT NULL,
 `some_field_2` TEXT NULL DEFAULT NULL
) 
```

**Our query**

```sql
-- @create
INSERT INTO some_table (
  some_field,
  some_field_2
) VALUES (
  @some_field,
  case when @some_field = 'TEST' then 'A' else NULL end
);
```

And in this case it generates:

```OCaml
  let create db ~some_field =
    let set_params stmt =
      let p = T.start_params stmt (19) in
      begin match some_field with None -> T.set_param_null p | Some v -> T.set_param_Text p v end;
      T.set_param_Text p some_field; 
      T.finish_params p
    in
    T.execute db ("INSERT INTO some_table (\n\
 some_field,
  some_field_2
) VALUES (
  ?,\n\
  ?,\n\
        case when ? like 'TEST' then 'A' else NULL end,\n\
)") set_params
```
This PR fixes it.

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -51,8 +51,12 @@ struct
   let common_subtype = common_type fst
   let common_type x y = Option.is_some @@ common_subtype x y
 
-  type tyvar = Typ of t | Var of int
-  let string_of_tyvar = function Typ t -> to_string t | Var i -> sprintf "'%c" (Char.chr @@ Char.code 'a' + i)
+  type tyvar = Typ of t | Var of int | Nulable of tyvar
+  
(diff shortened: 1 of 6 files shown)


Khady:
could you please add a test?

PR #121, opened 2023-12-07
Tracing-via-traits

ELLIOTTCABLE:
This PR implements feedback on #117, making tracing-information non-optional (and thus requiring no build-system changes), instead requiring traits-implementations to be updated to accept the new tracing-information.

This is a **breaking change** for custom traits-implementations.

The built-in traits implementation use [ocaml-trace](https://github.com/c-cube/ocaml-trace), if present in the installed world (`opam install trace`), to generate tracing-spans for each SQL statement.

All of the additions are noops at runtime if the user does not have `ocaml-trace` installed, and doesn't use a custom traits implementation.

### DIFF
--- a/impl/ocaml/mariadb/sqlgg_mariadb.ml
+++ b/impl/ocaml/mariadb/sqlgg_mariadb.ml
@@ -262,7 +262,8 @@ let no_params stmt =
   M.Stmt.execute stmt [||] >>=
   check
 
-let with_stmt db sql f =
+let with_stmt ?operation ?tables ~sql span_name db f =
+  Possibly_tracing.span ?operation ?tables ~system:`mariadb ~sql span_name @@ fun () ->
   let open IO in
   let close stmt = M.Stmt.close stmt >>= fun _ -> return () in
   M.prepare db sql >>=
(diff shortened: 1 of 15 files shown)


impl/ocaml/mariadb/sqlgg_mariadb.ml
```diff
@@ -271,8 +272,8 @@ let with_stmt db sql f =
 
 let row_array = (module M.Row.Array : M.Row.S with type t = M.Field.t array)
 
-let select db sql set_params callback =
-  with_stmt db sql @@ fun stmt ->
+let select ?operation ?tables ~span_name db sql set_params callback =
```
ygrek, line 275: lets put all meta information into one record `{ tables; span_name; operation }`

impl/ocaml/possibly_tracing.ml
```diff
@@ -0,0 +1,25 @@
...
+      | `mariadb -> "mariadb"
+      | `mysql -> "mysql"
+      | `sqlite -> "sqlite"
+    in
+    somes_into_strings
+      [
+        "db.system", Some system;
+        "db.statement", Some sql;
+        "db.operation", operation;
+        "db.table", Option.map List.hd tables;
```
ygrek, line 21: List.hd :sideeye:

ygrek (changes requested):
+ need to finish dune conditional

PR #123, opened 2023-12-13
Add JOIN statement tests

jongleb:
#### These test shows required improvements to make JOINS work properly.

When dealing with a LEFT JOIN, nullability arises under the following scenarios:

1. **No Foreign Key:**
   - If there is no foreign key relationship between the tables involved in the LEFT JOIN, nullability can occur when there is no matching record in the right table. In this case, all fields from the right table will be NULL.

2. **Nullable Foreign Key:**
   - Nullability may also arise when there is a nullable foreign key, and there is no matching record in the right table. If the foreign key in the left table allows NULL values, fields from the right table will be NULL in case of no match.

These scenarios highlight the conditions under which NULL values may appear in the result set of a LEFT JOIN operation.

### DIFF
--- a/src/test.ml
+++ b/src/test.ml
@@ -188,15 +188,9 @@ let test_manual_param = [
 
 let test_left_join_without_foreign_key_and_is_not_null = [
   tt "CREATE TABLE account_types ( type_id INT PRIMARY KEY, type_name VARCHAR(255) )" [] [];
-  tt "CREATE TABLE users ( user_id INT PRIMARY KEY, name VARCHAR(255), email VARCHAR(255), account_type_id INT )" [] [];
-   "LEFT JOIN" >::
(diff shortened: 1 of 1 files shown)


ygrek:
* for the first case the reason is in https://github.com/ygrek/sqlgg/blob/null/lib/syntax.ml#L330 (it works with just `type_name` without referencing table)
* for the second case it mentions `ON DELETE CASCADE`  in description, but not in the test

overall FOREIGN KEY was not handled until now at all, so thanks for bringing this up, but please uncomment the test cases so that they are actually failing

jongleb:
> lease uncomment the test cases so that they are actually failing

Done.

> for the second case it mentions ON DELETE CASCADE in description, but not in the 

I removed this option, thanks, I found out that nullability on JOIN depends only on the column nullability and on the reference existence

PR #128, opened 2024-02-28
Improve Coalesce Function for Proper Type Inference

jongleb:
This pull request enhances the coalesce function to provide accurate type inference. Previously, coalesce didn't behave as expected, but with this update, it correctly outputs a not null type if any of the parameters are not null.


Khady:
nitpick, but @jongleb I think you can get commit messages that are a bit more descriptive than just "coalesce"

PR #133, opened 2024-06-07
Nullability of tuple elements

jongleb:
This PR makes gen to consider the inferred nullability of types of tuple elements


jongleb:
https://github.com/ygrek/sqlgg/commit/d4bb359da5bf15646d5a1ce31230ce9e7e2e89c1

PR #140, opened 2024-10-22
add insert_id to sqlgg_mariadb execute response

tatchi:
add support for https://github.com/ocaml-community/ocaml-mariadb/pull/58

### DIFF
--- a/impl/ocaml/mariadb/sqlgg_mariadb.ml
+++ b/impl/ocaml/mariadb/sqlgg_mariadb.ml
@@ -198,7 +198,6 @@ type 'a connection = M.t
 type params = statement * M.Field.value array * int ref
 type row = M.Field.t array
 type result = M.Res.t
-type execute_response = { affected_rows : int; insert_id: int }
 
 module Types = Types
 
@@ -289,7 +288,7 @@ let execute db sql set_params =
   with_stmt db sql @@ fun stmt ->
   let open IO in
   set_params stmt >>=
-  fun res -> return { affected_rows = M.Res.affected_rows res; insert_id = M.Res.insert_id res }
+  fun res -> return { Sqlgg_traits.affected_rows = Int64.of_int (M.Res.affected_rows res); insert_id = Int64.of_int (M.Res.insert_id res) }
 
 
 let select_one_maybe db sql set_params convert =
--- a/impl/ocaml/mariadb/sqlgg_mariadb.ml
+++ b/impl/ocaml/mariadb/sqlgg_mariadb.ml
@@ -198,6 +198,7 @@ type 'a connection = M.t
 type params = statement * M.Field.value array * int ref
 type row = M.Field.t array
 type result = M.Res.t
+type execute_response = { affected_rows : int64; insert_id: int64}
 
 module Types = Types
 
@@ -288,7 +289,8 @@ let execute db sql set_params =
   with_stmt db sql @@ fun stmt ->
   let open IO in
   set_params stmt >>=
-  fun res -> return (Int64.of_int (M.Res.affected_rows res))
+  fun res -> return { affected_rows = Int64.of_int (M.Res.affected_rows res); insert_id = Int64.of_int (M.Res.insert_id res) }
+
 
 let select_one_maybe db sql set_params convert =
   with_stmt db sql @@ fun stmt ->
--- a/impl/ocaml/mariadb/sqlgg_mariadb.ml
+++ b/impl/ocaml/mariadb/sqlgg_mariadb.ml
@@ -198,7 +198,7 @@ type 'a connection = M.t
 type params = statement * M.Field.value array * int ref
 type row = M.Field.t array
 type result = M.Res.t
-type execute_response = { affected_rows : int64; insert_id: int64}
+type execute_response = { affected_rows : int64; insert_id: int }
 
 module Types = Types
 
@@ -289,7 +289,7 @@ let execute db sql set_params =
   with_stmt db sql @@ fun stmt ->
   let open IO in
   set_params stmt >>=
-  fun res -> return { affected_rows = Int64.of_int (M.Res.affected_rows res); insert_id = Int64.of_int (M.Res.insert_id res) }
+  fun res -> return { affected_rows = Int64.of_int (M.Res.affected_rows res); insert_id = M.Res.insert_id res }
 
 
 let select_one_maybe db sql set_params convert =
--- a/impl/ocaml/mariadb/sqlgg_mariadb.ml
+++ b/impl/ocaml/mariadb/sqlgg_mariadb.ml
@@ -198,6 +198,7 @@ type 'a connection = M.t
 type params = statement * M.Field.value array * int ref
 type row = M.Field.t array
 type result = M.Res.t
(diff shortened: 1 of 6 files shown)


sqlgg.opam
```diff
@@ -33,3 +33,7 @@ It starts off with SQL schema and queries, and generates code (or XML, allowing
 further code generation for various purposes). Generated code only defines a mapping
 of output columns and query parameters to the host language, trying to be as unobtrusive
 as possible and leaving the choice of SQL database (and API to access it) to the developer."""
+
+conflicts: [
+  "mariadb" {<= "1.1.6"}
+]
```
tatchi, line 39: This won't work with the current deployed versions of `mariadb`

impl/ocaml/mysql/sqlgg_mysql.ml
```diff
@@ -123,6 +123,7 @@ type 'a connection = Mysql.dbd
 type params = statement * string array * int ref
 type row = string option array
 type result = P.stmt_result
+type execute_response = int64
```
tatchi, line 126: open to any better name

impl/ocaml/mariadb/sqlgg_mariadb.ml
```diff
@@ -198,6 +198,7 @@ type 'a connection = M.t
 type params = statement * M.Field.value array * int ref
 type row = M.Field.t array
 type result = M.Res.t
+type execute_response = { affected_rows : int64; insert_id: int64}
```
tatchi, line 201: open to any better suggestion for this name

impl/ocaml/sqlgg_traits.ml
```diff
@@ -101,9 +102,9 @@ module type M = sig
 
   (** Execute non-query.
     @raise Oops on error
-    @return number of affected rows
+    @return execute_response
```
rr0gi, line 105: either explain what is execute_response or remove the line (it is clear from the type that it returns `execute_response`)

impl/ocaml/mariadb/sqlgg_mariadb.ml
```diff
@@ -288,7 +289,8 @@ let execute db sql set_params =
   with_stmt db sql @@ fun stmt ->
   let open IO in
   set_params stmt >>=
-  fun res -> return (Int64.of_int (M.Res.affected_rows res))
+  fun res -> return { affected_rows = Int64.of_int (M.Res.affected_rows res); insert_id = Int64.of_int (M.Res.insert_id res) }
```
rr0gi, line 292: why do Int64.of_int? it is less convenient type for no benefit

sqlgg.opam
rr0gi: with the current opam packaging this is kind of unfortunate - it means people who are using only sqlite traits will still have to deal with this conflict, ideally traits should be each separate opam lib 
i guess ok for this PR :/

rr0gi (changes requested):

impl/ocaml/sqlgg_traits.ml
tatchi: removed

impl/ocaml/mariadb/sqlgg_mariadb.ml
tatchi: This is just me blindly copying the current code 🙈. I switched `insert_id` to `int` in [`12ea564` (#140)](https://github.com/ygrek/sqlgg/pull/140/commits/12ea5649f79d4fc5445a790fd7b150a8f02ea588)

I got curious about `affected_rows`, it was added 15 years ago in https://github.com/ygrek/sqlgg/commit/c9e8d3f33380ea1ad8a92e72c9b8ae9ebaee4d3c#diff-6d600b7526702e1d382e1022fdf50144080017d3da236b6c4dc96b33529bb19b. At that time (and before this PR), there was no way to use a different `execute_response` type for each implementation and [`mysql` return `int64`](https://github.com/ygrek/ocaml-mysql/blob/master/mysql.mli#L247) so I believe that's the reason.

Since we can now specify the `execute_response` by implementation, I changed it to be `int` for `mariadb` and `sqlite`. It's a breaking change, but since there was one already I guess that's fine

tatchi: And now I'm realizing that `mysql` has also binding for [insert_id](https://github.com/ygrek/ocaml-mysql/blob/master/mysql.mli#L252). I adapted the `execute_response` type to also include `insert_id` [`6793f79` (#140)](https://github.com/ygrek/sqlgg/pull/140/commits/6793f792aa72f58c36dd60c38df68b3fd709c7f7)

tatchi: I'm confused about the difference between `mysql` and `mariadb`; both bind to `mysql_stmt_affected_rows` 🤔 Looks like there's a difference between mysql version? https://dev.mysql.com/doc/c-api/8.4/en/mysql-stmt-affected-rows.html vs https://dev.mysql.com/doc/c-api/5.7/en/mysql-stmt-affected-rows.html

tatchi: Ok looks like they're probably the same after all

> * my_ulonglong: This is a MySQL-specific typedef, which has been used traditionally in older versions (like 5.7) to represent 64-bit unsigned integers.
> * uint64_t: This is a standard C typedef introduced by the C99 standard, representing a 64-bit unsigned integer and is more widely recognized outside of MySQL.

But why `mysql` and `mariadb` use a different `OCaml` type. Which one is correct? Is it related to https://github.com/ocaml-community/ocaml-mariadb/issues/31 ?

tatchi:
So it looks like it's not that simple. I hit this error in a generated code by `sqlgg` with `mariadb` and the changes in this PR.

```ocaml
let add_many_person db ~persons =
  ( match persons with [] -> IO.return 0L | _ :: _ -> T.execute db ....
  
Error: This expression has type T.execute_response IO.future
       but an expression was expected of type int64 IO.future
       Type T.execute_response is not compatible with type int64 
```

I changed `OL` to the new `execute_response`, i.e `{ affected_rows = 0; insert_id = 0 }` but the problem is that it doesn't work for e.g `sqlite` which expects `execute_response` to be an `int64`.

I guess if we want to support different `execute_response` we'd have to expose a way to construct such response in the `Traits` (not entirely sure it'd work?).

Fortunately, I realized that we could also get the `insert_id` with `sqlite`, so instead we could just have all implementations to return the same type (as currently). That's what I did in [`e5e9626` (#140)](https://github.com/ygrek/sqlgg/pull/140/commits/e5e9626e9dbcfdcd8cbba4db20c92ab22266c1ae). You can see that I use `int64` because there's always at least one implementation that requires it, and we need to use a common type. This is probably also why we currently have `int64` and not `int`.

impl/ocaml/mariadb/sqlgg_mariadb.ml
rr0gi: > But why mysql and mariadb use a different OCaml type. Which one is correct? 

it is different libraries, they made different choices at some point, correct depends on how you define correct :)

PR #142, opened 2024-11-04
Introduce shortcut syntax for optional bool in `WHERE` expr

jongleb:
This PR adds new syntaxes:

1) `@param :: Type?`
2) `WHERE A? AND B? OR C? which is a shortcut to WHERE @choice { None { TRUE } | Some (params) { A } } AND etc` 
3) `WHERE A! AND B! OR C! which expands to WHERE @choice { None { FALSE } | Some { A } } AND etc`

And considers `join_cond` as not null by the default 

Example: 

DDL: 

```sql
CREATE TABLE test19 (
  a INT NOT NULL,
  b INT NOT NULL
);
CREATE TABLE test20 (
  c INT NOT NULL,
  d INT NOT NULL,
  r TEXT NOT NULL
);
```

SQL: 
```sql
SELECT *
FROM test19
LEFT JOIN test20 on test20.c = @test20a
WHERE { c = @choice2 }? AND { r = @choice3 }! AND r = @var3
GROUP BY b;
```
will be generated the following code: 

```ocaml
let test db ~test20a ~choice2 ~choice3 ~var3 callback =
    let invoke_callback stmt =
      callback
        ~a:(T.get_column_Int stmt 0)
        ~b:(T.get_column_Int stmt 1)
        ~c:(T.get_column_Int_nullable stmt 2)
        ~d:(T.get_column_Int_nullable stmt 3)
        ~r:(T.get_column_Text_nullable stmt 4)
    in
    let set_params stmt =
      let p = T.start_params stmt (4) in
      T.set_param_Int p test20a;
      begin match choice2 with
      | None -> ()
      | Some (choice2) ->
        T.set_param_Int p choice2;
      end;
      begin match choice3 with
      | None -> ()
      | Some (choice3) ->
        T.set_param_Text p choice3;
      end;
      T.set_param_Text p var3;
      T.finish_params p
    in
    T.select db ("SELECT *\n\
FROM test19\n\
LEFT JOIN test20 on test20.c = ?\n\
WHERE " ^ (match choice2 with Some _ -> " c = " ^ "?" ^ " " | None -> " TRUE ") ^ " AND " ^ (match choice3 with Some _ -> " r = " ^ "?" ^ " " | None -> " FALSE ") ^ " AND r = ?\n\
GROUP BY b") set_params invoke_callback
```

Also, it differs from the regular `Choice` in that, in this case, not a polymorphic variant is generated because for this case, it would lead to the useless matching from the regular `option` type to the poly version of the option in the user code.

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -340,6 +340,7 @@ let show_table_name { db; tn } = match db with Some db -> sprintf "%s.%s" db tn 
 let make_table_name ?db tn = { db; tn }
 type schema = Schema.t [@@deriving show]
 type table = table_name * schema [@@deriving show]
+type pos = (int * int) [@@deriving show]
 
 let print_table out (name,schema) =
   IO.write_line out (show_table_name name);
@@ -349,7 +350,7 @@ let print_table out (name,schema) =
   IO.write_line out ""
 
 (** optional name and start/end position in string *)
-type param_id = { label : string option; pos : int * int; } [@@deriving show]
+type param_id = { label : string option; pos : pos; } [@@deriving show]
 type param = { id : param_id; typ : Type.t; } [@@deriving show]
 let new_param id typ = { id; typ; }
 type params = param list [@@deriving show]
@@ -362,7 +363,7 @@ and var =
 | ChoiceIn of { param: param_id; kind : [`In | `NotIn]; vars: var list }
 | Choice of param_id * ctor list
 | TupleList of param_id * tuple_list_kind
-| OptionBoolChoice of bool * param_id * var list * (int * int)
+| OptionBoolChoice of bool * param_id * var list * (pos * pos)
 and tuple_list_kind = Insertion of schema | Where_in of Type.t list
 [@@deriving show]
 type vars = var list [@@deriving show]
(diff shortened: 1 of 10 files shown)


ygrek:
didn't look at the code

1. why is `:: Type?` here?
2. So it means if A? OR B? and both are None - then condition is true and everything is selected.
3. What is expected usage ratio for A! and A? One obviously can be expressed with the other (or via current manual unwrapping), so if one is very low use - i would prefer to not introduce extra syntax.

jongleb:
> why is :: Type? here?

Just shortcut
Remove?

> So it means if A? OR B? and both are None - then condition is true and everything is selected.

Yes, something like with `COALESCE` (maybe a bad analogy), while the first is not null - type is always null

> What is expected usage ratio for A! and A? One obviously can be expressed with the other (or via current manual unwrapping), so if one is very low use - i would prefer to not introduce extra syntax.

Ok I'll remove it

ygrek:
ok, so keep this PR only about introducing automatic elimination for A? OR B?
I am also slightly uneasy about `::Type? ` precisely because it means different thing from `WHERE A?` which may be possibly confusing? Maybe it is worth to be more explicit with :: Maybe Type or some such or maybe I am overthinking

jongleb:
@ygrek I removed the unnecessary stuff, made a couple of edits and allowed myself to merge, since this PR is already a month old, if there will any new comments, I will fix them in a separate PR

PR #143, opened 2024-11-05
Make subqueries nullable and consider count 

jongleb:
### Description

This PR adds proper support of nullability inferring of subqueries

Examples:

DDL:
```sql
 CREATE TABLE table_30 (
      column_1 INT PRIMARY KEY,
      column_2 VARCHAR(50) NOT NULL,
      column_3 VARCHAR(50)
    )
```

Query                       | Before PR    | Since PR 
|-------------------|-------------|------------|
| ```SELECT (SELECT 1 WHERE 0) as result ```  | Not null |  Null 
| ```SELECT 1 as one, (SELECT COUNT(NULL)) as result```  | Not null |  Not Null
| ``` SELECT 'test' as t, (SELECT column_1 FROM table_30 WHERE FALSE LIMIT 1)```  | Not null | Null
| ```SELECT 1 as one, (SELECT COUNT(NULL) HAVING FALSE) as result``` | Not null | Null

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -110,7 +110,9 @@ struct
   type tyvar = Typ of t | Var of int
   let string_of_tyvar = function Typ t -> show t | Var i -> sprintf "'%c" (Char.chr @@ Char.code 'a' + i)
 
-  type agg_fun = Self | Count | Avg (* max, min | count | avg *)
+  type agg_fun = Self (* self means that it returns the same type what aggregated columns have. ie: max, min *) 
+    | Count (* count it's count function which never returns null  *) 
+    | Avg (* avg it's avg function that returns float *)
 
   type func =
   | Agg of agg_fun (* 'a -> 'a | 'a -> t *)
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -110,9 +110,11 @@ struct
   type tyvar = Typ of t | Var of int
   let string_of_tyvar = function Typ t -> show t | Var i -> sprintf "'%c" (Char.chr @@ Char.code 'a' + i)
 
+  type agg_fun = Self | Compute of t (* max, min | avg *)
+
   type func =
-  | Group of t (* _ -> t *)
-  | Agg (* 'a -> 'a *)
+  | Count (* _ -> t *)
+  | Agg of agg_fun (* 'a -> 'a | 'a -> t *)
   | Multi of tyvar * tyvar (* 'a -> ... -> 'a -> 'b *)
   | Coalesce of tyvar * tyvar
   | Comparison
@@ -127,8 +129,9 @@ struct
   let pp_func pp =
     let open Format in
   function
-  | Agg -> fprintf pp "|'a| -> 'a"
-  | Group ret -> fprintf pp "|_| -> %s" (show ret)
+  | Agg Self -> fprintf pp "|'a| -> 'a"
+  | Agg Compute t -> fprintf pp "|'a| -> %s" (show t)
+  | Count -> fprintf pp "[_] -> int"
   | Ret ret -> fprintf pp "_ -> %s" (show ret)
   | F (ret, args) -> fprintf pp "%s -> %s" (String.concat " -> " @@ List.map string_of_tyvar args) (string_of_tyvar ret)
   | Multi (ret, each_arg) | Coalesce (ret, each_arg) -> fprintf pp "{ %s }+ -> %s" (string_of_tyvar each_arg) (string_of_tyvar ret)
@@ -137,7 +140,7 @@ struct
   let string_of_func = Format.asprintf "%a" pp_func
 
   let is_grouping = function
-  | Group _ | Agg -> true
+  | Count | Agg _ -> true
   | Ret _ | F _ | Multi _ | Coalesce _  | Comparison -> false
 end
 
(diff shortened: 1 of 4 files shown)


src/test.ml
```diff
@@ -569,6 +569,165 @@ let test_ambiguous = [
...
+  |} [
+    attr' ~extra:[NotNull] "info1" Text;
+    attr' ~nullability:Nullable "info2" Text;
+    attr' ~nullability:Nullable "max_info" Int;
+  ] [];
+
+  tt {| 
+    SELECT (SELECT 1 WHERE 0) as result
+  |} [
+    attr' ~nullability:Nullable  "result" Int;
```
ygrek, line 654: (not for this PR) need better way to describe test results, this is such eyesore to read when it should be like `int ~nullable "result"` (with `let nullable = Nullable` above in scope) or some such

lib/sql.ml
```diff
@@ -110,13 +110,14 @@ struct
   type tyvar = Typ of t | Var of int
   let string_of_tyvar = function Typ t -> show t | Var i -> sprintf "'%c" (Char.chr @@ Char.code 'a' + i)
 
+  type agg_fun = Self | Count | Avg (* max, min | count | avg *)
```
ygrek, line 113: i was thinking i would prefer to have return type here (`Self | Return of strict int | Return of nullable float), but then realize that indeed it is count and avg are so special and if there are no other functions that behave in same way then this also makes sense. But please put comment that describes why these functions are special

lib/syntax.ml
```diff
@@ -255,9 +255,27 @@ let rec resolve_columns env expr =
...
       match schema, usage with
-      | [ {domain;_} ], `AsValue -> ResFun (Type.Ret domain.t, as_params p) (* use nullable *)
+      | [ {domain;_} ], `AsValue ->
+        (* This function should be raised? *)
+        let rec with_count = function 
+          | Fun (Agg Count, _)
+          | SelectExpr (_, _) -> Some domain
+          | Fun (_, exprs) -> List.find_map with_count exprs
+          | Value _| Param _| Inparam _| Choices (_, _) | InChoice (_, _, _)
+          | Column _| Inserted _| InTupleList (_, _) -> None
```
ygrek, line 265: Choices should recurse, not None?

lib/syntax.ml
```diff
@@ -255,9 +255,27 @@ let rec resolve_columns env expr =
...
+        in
+        let default_null = Type.make_nullable domain in
+        (* The only way to have a result in a subquery is to use the count function. 
+           Any other expression could possibly return no rows. *)
+        let typ = match select.select_complete.select with 
+        | ({ having = Some _; _ }, _) -> Type.nullable domain.t
```
ygrek, line 271: need comment that HAVING still makes it nullable even with COUNT

lib/syntax.ml
```diff
@@ -255,9 +255,27 @@ let rec resolve_columns env expr =
...
+        | ({ columns = [Expr(c, _)]; _ }, _) -> c |> with_count |> Option.default default_null
+        | ({ columns = [_]; _ }, _) -> default_null
+        | _ -> raise (Schema.Error (schema, "Different count of selected columns are presented"))
```
ygrek, line 274: this error wouldn't make sense to me without looking at the code. Should say smth like "nested sub-select used as an expression returns more than one column"

lib/syntax.ml
```diff
@@ -255,9 +255,27 @@ let rec resolve_columns env expr =
...
+        in
+        ResFun (Type.Ret (typ), as_params p) (* use nullable *)
```
ygrek, line 276:
```suggestion
        ResFun (Type.Ret (typ), as_params p)
```
no longer needed because implemented here i understand?

ygrek (approved):
lgtm, some comments about comments :doge:

lib/syntax.ml
cyberhuman: BTW I think it's possible in MySQL 8 to return the row result.

src/test.ml
jongleb: Got it

lib/sql.ml
jongleb: Hmm... hm hm hm, I am thinking now, and I realize that you are right about `avg`.
`Count` - should be kept, no one aggregation in the MySQL world can't be non-nullable and every-time-row-returnable, unlike count.
When I was adding `avg,` my thoughts were that we split functions for the following kinds: self, which means it returns the same type as rows have, and count (already described), and avg, which returns not self type but produces calculation and return a float.
But now I chatGPTed about which functions do calculations besides avg, and it returned to me: `STD` (I checked it and , yes, it has the same type that avg),  `VAR_POP` (same),  `GROUP_CONCAT` - okay it deserves a new constructor, `BIT_AND` (it accepts   `TINYINT`, `SMALLINT`, `MEDIUMINT`, `INT`, `BIGINT` and returns `BIGINT`, and how it should be described. Roughly: Int -> Int Nullable.

jongleb: `type agg_fun = Self | Count | Return of int | Return of floar`

jongleb: Another problem, is that it's possible to write:

```sql
SELECT AVG('string');
```

It won't raise an error in runtime. it will try to cast a string to int, but I guess it is better to restrict it.
And the second question is where to restrict?
To return something like during inferring: 🤔

```ocaml
 | Agg Avg, [_] -> 
          let (_, typ) = typeof (ResFun ((F ((Typ (nullable Float)), [(Typ (depends Int))])), params)) in 
          let typ = get_or_failwith typ in
          consider_groupping @@ typ, types
```
But this way we miss Float

jongleb: > But this way we miss Float

But we can match argument, okay better I guess to to "desugaring" carefully before calling infer.

jongleb: Okay, for now, I will check it in pattern matching during infer, but what if we will do something like typeclasses
https://github.com/andreypopp/type-systems/blob/main/algo_w/infer.ml#L8 to pick Number class, but yes not for this PR

lib/syntax.ml
jongleb: Yes, thanks

lib/sql.ml
ygrek: > It won't raise an error in runtime. it will try to cast a string to int, but I guess it is better to restrict it.

currently sqlgg doesn't do any validation and it is good for adoption. Validation will require better impl for type inference etc, and anyhow must be behind the switch to not disrupt

ygrek (approved):

PR #144, opened 2024-11-20
Fix WHERE IN inside subquery

jongleb:
This PR fixes `WHERE IN` expressions inside subqueries

### DIFF
--- a/test/out/where_in_in_subquery.xml
+++ b/test/out/where_in_in_subquery.xml
@@ -1,0 +1,110 @@
+<?xml version="1.0"?>
+
+<sqlgg>
+ <stmt name="create_test30" sql="CREATE TABLE test30 (&#x0A;    id INT PRIMARY KEY, &#x0A;    column_a TEXT, &#x0A;    column_b BOOL, &#x0A;    column_c_31 INT NOT NULL&#x0A;)" category="DDL" kind="create" target="test30" cardinality="0">
+  <in/>
+  <out/>
+ </stmt>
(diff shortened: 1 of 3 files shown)


tatchi:
Thanks for the fix 🙏 Would it make sense to create a test for it ?

test/out/where_in_in_subquery.xml
```diff
@@ -0,0 +1,110 @@
...
+  <in/>
+  <out/>
+ </stmt>
+ <stmt name="create_test32" sql="CREATE TABLE test32 (&#x0A;    id INT, &#x0A;    column_f INT&#x0A;)" category="DDL" kind="create" target="test32" cardinality="0">
+  <in/>
+  <out/>
+ </stmt>
+ <stmt name="update_3" sql="UPDATE test30 t30&#x0A;SET&#x0A;    t30.column_a = 'the value',&#x0A;    t30.column_b = false&#x0A;WHERE&#x0A;    t30.column_c_31 IN (&#x0A;        SELECT t31.id&#x0A;        FROM test31 t31&#x0A;        WHERE t31.column_c_32 IN (&#x0A;            SELECT t32.id&#x0A;            FROM test32 t32&#x0A;            WHERE t32.column_f IN @@c_f_ids&#x0A;        )&#x0A;    )" category="DML" kind="update" cardinality="0">
+  <in>
+   <value name="c_f_ids" type="set(Int)" nullable="true"/>
```
jongleb, line 18: It's actually generates list, not a set, but we understand that it reflects that we wouldn't want to have duplications inside `WHERE IN (..)` expression

PR #148, opened 2025-01-20
implement values row syntax , first part (joins)

jongleb:
### Description
This PR adds a support of `VALUES ROW(..)` syntax.

https://dev.mysql.com/doc/refman/8.4/en/values.html

At the current step, it's supported for `JOIN`S, but it's already now considered as a new source, so it wouldn't be a problem to support the rest of the usages of this syntax.

Also it supports two variants of it:

```
SELECT p.id, x.a
FROM products p
JOIN ( VALUES ROW(...) ) AS x (a, b)
ON p.name = x.b;
```

And

```
SELECT p.id, x.a
FROM products p
JOIN ( VALUES @param ) AS x (a, b)
ON p.name = x.b;
```

And since `VALUES` with no ROWs afters is invalid syntax, I added a default expression when the passed list is empty:

```SELECT %cols% WHERE FALSE```

This part is up for debate, but it seems safer

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -365,7 +365,7 @@ and var =
 | TupleList of param_id * tuple_list_kind
 (* It differs from Choice that in this case we should generate sql "TRUE", it doesn't seem reusable *)
 | OptionBoolChoice of param_id * var list * (pos * pos)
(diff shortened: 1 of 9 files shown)


jongleb:
@ygrek Could you take a look , please

PR #149, opened 2025-01-21
Make execute_response's insert_id : int64 option

andreypopp:
To represent explicitly the case when the insert wasn't performed (can be because of `ON CONFLICT` clause for example).

### DIFF
--- a/impl/ocaml/sqlgg_traits.ml
+++ b/impl/ocaml/sqlgg_traits.ml
@@ -27,7 +27,7 @@ module type M = sig
   type params
   type row
   type result
-  type execute_response = { affected_rows: int64; insert_id: int64 }
+  type execute_response = { affected_rows: int64; insert_id: int64 option }
 
   (** datatypes *)
   module Types : sig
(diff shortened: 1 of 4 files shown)


impl/ocaml/sqlgg_traits.ml
```diff
@@ -27,7 +27,7 @@ module type M = sig
   type params
   type row
   type result
-  type execute_response = { affected_rows: int64; insert_id: int64 }
+  type execute_response = { affected_rows: int64; insert_id: int64 option }
```
tatchi, line 30: Should we document that if `insert_id` is `Some x`, `x` is guaranteed to be greater than 0?

tatchi: Also, I wonder if we can have `insert_id > 0` and `affected_rows = 0` (or the other way around). If not, maybe it makes sense to have instead

```ocaml
type execute_response = { affected_rows: int64; insert_id: int64 } option
```

Where both `affected_rows` and `insert_id` are guaranteed to be greater than 0 🤔

jongleb: affected_rows > 0 and insert_id = None When UPDATE or DELETE.

PR #152, opened 2025-02-07
Implement safe enums: literals + poly variants gen

jongleb:
### Description
This PR introduces two main enums related features:

#### 1. String Literal Types and Unions
The primary feature is that any string literal now has a StringLiteral type. When used in expressions like:

```
ROW..VALUES
SELECT .. UNION
WHERE IN (...)
@param { ...} | { ... } | {...}
```

or any other cases where `StringLiteral` can accept multiple values, the type will be inferred as a Union.
There are two types of Unions:

Open Union (resulting from combining StringLiterals or other Unions)
Closed Union (literal/Enum)

The relationships between these types can be seen [here](https://github.com/ygrek/sqlgg/pull/152/files#diff-080579952fb3ca76eab28b5486112bfe26d6ba34f8f0edcd115f0a37aee1a168R78-R105)

Also check [tests](https://github.com/ygrek/sqlgg/pull/152/files#diff-8ea9ba782f39d6a4f13234e5c57b9662624994b402b68b3b1e62bd255f7e10b5R821-R883)

This PR adds a new flag `type-safe-enums` (name is negotiable). When this flag is enabled parameters to enum fields and enum literals (new syntax too) now are checked.

##### Type Inference and Subtyping Example

```sql
INSERT INTO `tblname` (`col1`, `col2`, `col3`, `col4`, `col5`) 
SELECT @param3 { A { 'A' } | B { 'B' } | C { 'C' } }
-- rest of the query omitted for clarity
```

##### Enum Type Inference
In this case, the matching values are inferred as:

StringLiteral 'A'
StringLiteral 'B'
StringLiteral 'C'

These are then combined into an enum type A | B | C. The system verifies that this enum is either the same type as or a subtype of the closed enum A | B | C

##### Subtyping Rules
For example, this would be valid:

```sql
@param3 { A { 'A' } | B { 'B' } }
```

Because A | B is a valid subtype of A | B | C - you can safely write a value of type A | B where A | B | C is expected.

##### String Operations and Type Coercion
String operations work naturally:

```sql
SELECT CONCAT(col3, 'test') FROM tblname;
```

This is valid because Enum is a subtype of Text, so the enum value is automatically coerced to text. The reverse operation (text to enum) is not allowed.
Here's a complete example demonstrating insertion with string concatenation:

```sql
INSERT INTO tblname (col3) 
SELECT CONCAT(
  @param3 { A { 'A' } | B { 'B' } | C { 'C' } },
  '_suffix'
);
```

This will fail type checking because:

col3 expects an enum of type A | B | C
CONCAT() produces a Text type
You cannot assign a Text value to an enum field - even if the text matches one of the enum variants (like 'A', 'B', or 'C')

```sql
INSERT INTO tblname (col3) VALUES ('A');  -- Ok
INSERT INTO tblname (col3) VALUES (CONCAT('A', '')); -- Fails: Text to Enum
```

##### DDL:

```sql
CREATE TABLE `some_table` (
  `id` int(10) UNSIGNED NOT NULL AUTO_INCREMENT,
  `status` enum('pending','sending','sent','cancelled')  NOT NULL,
  `status_b` enum('a','b','c','d')  NOT NULL
);
```

#### Example (flag is disabled)

```sql
INSERT INTO `some_table` (
  `status`,
  `status_b`
) VALUES (
  'pending',
  @status_b
);
```

Here nothing is changed

```ocaml
  let insert_some_table_0 db ~status_b =
    let set_params stmt =
      T.set_param_Text p status_b;
      T.finish_params p
    in
    T.execute db ("INSERT INTO `some_table` (\n\
  `status`,\n\
  `status_b`\n\
) VALUES (\n\
  'pending',\n\
  ?\n\
)") set_params
```

This is what is generated

#### Example (flag is enabled)

```sql
INSERT INTO `some_table` (
  `status`,
  `status_b`
) VALUES (
 'pending',
  @status_b
);
```

This is the result

```ocaml
    module Enum_0 = T.Make_enum(struct
      type t = [`A | `B | `C | `D]
      let inj = function | "a" -> `A | "b" -> `B | "c" -> `C | "d" -> `D | s -> failwith (Printf.sprintf "Invalid enum value: %s" s)
      let proj = function  | `A -> "a"| `B -> "b"| `C -> "c"| `D -> "d"
    end)

  let insert_some_table_0 db ~status_b =
    let set_params stmt =
      Enum_0.set_param p status_b;
      T.finish_params p
    in
    T.execute db ("INSERT INTO `some_table` (\n\
  `status`,\n\
  `status_b`\n\
) VALUES (\n\
  'pending',\n\
  ?\n\
)") set_params
```

### DIFF
--- a/impl/ocaml/mariadb/sqlgg_mariadb.ml
+++ b/impl/ocaml/mariadb/sqlgg_mariadb.ml
@@ -24,6 +24,14 @@ module type Value = sig
   val to_literal : t -> string
 end
 
+module type Enum = sig 
+  type t
+
+  val inj: string -> t
+
+  val proj: t -> string
+end
+
 module type Types = sig
   type field
   type value
@@ -36,6 +44,7 @@ module type Types = sig
   module Datetime : Value
   module Decimal : Value
   module Any : Value
+  module Make_enum : functor (E : Enum) -> Value with type t = E.t
 end
 
 module Default_types(M : Mariadb.Nonblocking.S) : Types with
@@ -183,6 +192,20 @@ struct
     let to_value x = x
     let to_literal _ = failwith "to_literal Any"
   end)
+
+  module Make_enum (E: Enum) = Make(struct
+  
+    include E
+  
+    let of_field field = 
+      match M.Field.value field with
+      | `String x -> inj x
+      | value -> convfail "enum" field value
+  
+    let to_value v = `String (proj v)
+  
+    let to_literal = proj
+  end)
 end
 
 module Make
@@ -257,6 +280,19 @@ let set_param_Int = set_param_ty Int.to_value
 let set_param_Float = set_param_ty Float.to_value
 let set_param_Decimal = set_param_ty Decimal.to_value
 let set_param_Datetime = set_param_ty Datetime.to_value
+
+module Make_enum (E: Enum) = struct 
+
+  module E = Make_enum(E)
+
+  type t = E.t
+
+  let get_column, get_column_nullable = get_column_ty "Enum" E.of_field
+
+  let set_param = set_param_ty E.to_value
+
+  let to_literal = E.to_literal
+end
 
 let no_params stmt =
   let open IO in
--- a/impl/ocaml/mysql/sqlgg_mysql.ml
+++ b/impl/ocaml/mysql/sqlgg_mysql.ml
@@ -126,6 +126,15 @@ type result = P.stmt_result
 type execute_response = { affected_rows: int64; insert_id: int64 }
 
 module Types = T
+
+module type Enum = sig 
+  type t
+
+  val inj: string -> t
+
+  val proj: t -> string
+end
+
 open Types
 
 (* compatibility *)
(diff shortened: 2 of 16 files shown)


jongleb:
> Why need new syntax? we know from type that enum is expected.

because type inference is based on syntax itself. much of the simplicity of the hm for example type system is that it allows types to be automatically inferred from syntax. in this case when we don't have special syntax  we make our `Text` type more cunning and including the value in order to check the context of use at further stages and if the context is enum then only then consider it as enum value, I therefore decided to have a special syntax.

> Also I would consider dropping the flag, make this behaviour default and allow to opt-out on per-field basis.

got it

ygrek:
special syntax is definitely not a requirement by HM
it may be a requirement imposed by hacky implementation of something HM-like in sqlgg :)
but i expect it should be able to figure it out through unification same as all other variables/parameters where we don't know type beforehand

jongleb:
@ygrek I considered your comment, it now doesn't require writing special syntax for the user it's regular strings. I am going to merge it,  I will back and fix if there are any other comments

PR #154, opened 2025-02-11
remove compilation warning (deprecation, unlabbeled parameter)

raphael-proust:
Simple changes to remove some compilation warnings.

### DIFF
--- a/src/test.ml
+++ b/src/test.ml
@@ -142,7 +142,7 @@ let test_parsing = [
 let test_join_result_cols () =
   Tables.reset ();
   let ints = List.map (fun name ->
-    if String.ends_with name "?" then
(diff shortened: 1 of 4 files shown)


src/test.ml
```diff
@@ -142,7 +142,7 @@ let test_parsing = [
 let test_join_result_cols () =
   Tables.reset ();
   let ints = List.map (fun name ->
-    if String.ends_with name "?" then
+    if String.ends_with name ~suffix:"?" then
```
raphael-proust, line 145: especially confusing because `Stdlib.String` and `ExtString` have different argument order

ygrek (changes requested):
please remove src/clil.exe from PR and lgtm
thanks!

PR #156, opened 2025-02-12
add cli.exe to gitinore

raphael-proust:
uses promote in dune file so it shows up in the repo


ygrek:
committed similar change directly

PR #157, opened 2025-02-14
factorise functions block in traits

raphael-proust:
some light factorisation to avoid code-duplication

(because I'm about to open another PR which makes some changes to the factored-out block)

### DIFF
--- a/impl/ocaml/sqlgg_traits.ml
+++ b/impl/ocaml/sqlgg_traits.ml
@@ -34,12 +34,27 @@ module type FNS = sig
 
   val no_params : statement -> result io_future
 
+  (**
+    Perform query (cardinality "any") and return results via callback for each row
+    @raise Oops on error
+  *)
   val select : [>`RO] connection -> string -> (statement -> result io_future) -> (row -> unit) -> unit io_future
 
+  (**
+    Perform query (cardinality "zero or one") and return first row if available
+    @raise Oops on error
+  *)
   val select_one_maybe : [>`RO] connection -> string -> (statement -> result io_future) -> (row -> 'b) -> 'b option io_future
 
+  (**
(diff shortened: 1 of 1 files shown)


impl/ocaml/sqlgg_traits.ml
```diff
@@ -80,30 +102,14 @@ module type M = sig
...
-  (**
-    Perform query (cardinality "any") and return results via callback for each row
-    @raise Oops on error
-  *)
-  val select : [>`RO] connection -> string -> (statement -> result) -> (row -> unit) -> unit
-
-  (**
-    Perform query (cardinality "zero or one") and return first row if available
-    @raise Oops on error
-  *)
```
ygrek, line 94: doc comments got lost

impl/ocaml/sqlgg_traits.ml
```diff
@@ -113,16 +119,13 @@ module type M_io = sig
...
-
-  val execute : [>`WR] connection -> string -> (statement -> result IO.future) -> execute_response IO.future
+  include FNS
+    with type params := params
+    with type result := result
+    with type 'a io_future := 'a IO.future
+    with type 'a connection := 'a connection
+    with type statement := statement
+    with type row := row
+    with type execute_response := execute_response
```
ygrek, line 129: this slight visual sore prompts me to question why M_io is needed as separate module, instead of making IO-aware module a functor and then instantiate synchronous version with `type 'a future = 'a`, but not for this PR ofc

ygrek (changes requested):

impl/ocaml/sqlgg_traits.ml
raphael-proust: re-added the comments

ygrek:
thanks

PR #158, opened 2025-02-14
Specialise inserts

raphael-proust:
on top of https://github.com/ygrek/sqlgg/pull/157

This PR treats different kinds of non-query statements differently. Specifically, it specialises the function generated for `INSERT` statements to fail when `insert_id=0L` and sets a different return type for `INSERT` statements which have an `ON CONFLICT` clause (to strongly encourage users to check whether the statement inserted or not).

This is to restore some of the backwards compatibility with the state that the library was in before the introduciton of the option type in the return of `insert`.

### DIFF
--- a/src/gen_caml.ml
+++ b/src/gen_caml.ml
@@ -159,13 +159,23 @@ let output_select1_cb _ schema =
   output "in";
   name
 
+let has_on_conflict schema =
+  (List.exists (fun { extra; _ } -> Constraints.exists (function OnConflict _ -> true |
+    PrimaryKey | NotNull | Null | Unique | Autoincrement | WithDefault
+        -> false) extra) schema)
+
 let output_schema_binder index schema kind =
-  match schema with
-  | [] -> "execute",""
-  | _ -> match kind with
-         | Stmt.Select `Zero_one -> "select_one_maybe", output_select1_cb index schema
-         | Select `One -> "select_one", output_select1_cb index schema
-         | _ -> "select",output_schema_binder index schema
+  match schema, kind with
+  | [], Stmt.Insert (_, _) ->
+      if has_on_conflict schema then
+        "insert_with_on_conflict",""
+      else
+        "insert",""
+  | [], _ -> "execute",""
+  | _, Select `Zero_one -> "select_one_maybe", output_select1_cb index schema
+  | _, Select `One -> "select_one", output_select1_cb index schema
+  | _, _ -> "select",output_schema_binder index schema
+
 
 let is_callback stmt =
   match stmt.schema, stmt.kind with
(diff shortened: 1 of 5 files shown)


src/gen_caml.ml
```diff
@@ -159,13 +159,23 @@ let output_select1_cb _ schema =
   output "in";
   name
 
+let has_on_conflict schema =
+  (List.exists (fun { extra; _ } -> Constraints.exists (function OnConflict _ -> true |
```
raphael-proust, line 163: Should we match further on the payload of the `OnConflict` constructor?

`OnConflict Ignore` and `OnConflict Replace` could be treated differently: the former returning `true` the latter `false`?

raphael-proust: Also, maybe we sould also check `ON DUPLICATE KEY`?

jongleb: maybe, but do we really need to try to differentiate this and there is https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html check by the way

src/gen_caml.ml
```diff
@@ -489,7 +499,13 @@ let generate_stmt style index stmt =
     | Some id ->
     match id.label with
     | None -> failwith "empty label in tuple substitution"
-    | Some label -> sprintf {|( match %s with [] -> IO.return { T.affected_rows = 0L; insert_id = None } | _ :: _ -> %s)|} label exec
+    | Some label ->
+    match stmt.kind with
+    | Insert (_, _) when has_on_conflict stmt.schema ->
+        sprintf {|( match %s with [] -> IO.return { T.affected_rows = 0L; maybe_insert_id = None } | _ :: _ -> %s)|} label exec
+    | _ ->
+        sprintf {|( match %s with [] -> IO.return { T.affected_rows = 0L; insert_id = 0L } | _ :: _ -> %s)|} label exec
```
jongleb, line 507: in this case, if nothing is written, we get `oops insert has no insert_id` , right?

src/gen_caml.ml
raphael-proust: ultimately I don't know sql well enough to be sure of what we want to include in the check

it seems the addition of the `option` type was due to an `ON DUPLICATE KEY`, but as far as I can tell (which is not very far) the documentation of sql seems to suggest that `ON DUPLICATE KEY` should set the last-insert (to the modifed row)…

So essentially I don't know the domain of sql (the small difference between the different flavours, the extent of the language, all the different ways that an insert can suceed without data being inserted) to make a call here, I only know that the introduction of the `option` type breaks the interface a lot which makes it impractical to update.

src/gen_caml.ml
```diff
@@ -159,13 +159,23 @@ let output_select1_cb _ schema =
...
 let output_schema_binder index schema kind =
-  match schema with
-  | [] -> "execute",""
-  | _ -> match kind with
-         | Stmt.Select `Zero_one -> "select_one_maybe", output_select1_cb index schema
-         | Select `One -> "select_one", output_select1_cb index schema
-         | _ -> "select",output_schema_binder index schema
+  match schema, kind with
+  | [], Stmt.Insert (_, _) ->
+      if has_on_conflict schema then
```
jongleb, line 170: @ygrek, what do you say about this `has_on_conflict`? Should we keep it, and should we consider cases such as https://dev.mysql.com/doc/refman/8.4/en/create-table-check-constraints.html?
The rest part is okay and hasn't raised questions from my side.

jongleb:
@raphael-proust Please rebase, and I'll merge it

raphael-proust:
rebased

we need to work on the selection criteria for switching to the specialised option-based insert, but that can be done in a future PR

PR #159, opened 2025-03-18
Add subqueries ctor instead of ResFun

jongleb:
### Description

This PR fixes a bug related to subqueries and the Choices expression.

Problem
Before this PR, when inferring a subquery, we would extract vars, but then attempt to reconstruct an expression from them ([see this code block](https://github.com/ygrek/sqlgg/pull/159/files#diff-e5b2ac5b61488c46149361560003e3185b33725d73f243f2cce2fc3020757438L277-L290)). However, an exact isomorphism from res_expr to vars is not always possible, as many expressions cannot be reconstructed from vars.

For example, consider the following query:

sql
Копировать код
select 1 where 1 in (select 1 where @foo { Foo { IF (...) } | Bar { IF (...) } })
In this case, inference produces vars, but we cannot reconstruct the original Choices expression.

Solution
This PR removes the problematic behavior and introduces a new ResSelect constructor to store vars

### DIFF
--- a/lib/syntax.ml
+++ b/lib/syntax.ml
@@ -37,6 +37,7 @@ type enum_ctor_value_data = { ctor_name: string; pos: pos; }  [@@deriving show]
 type res_expr =
   | ResValue of Type.t (** literal value *)
   | ResParam of param
+  | ResSelect of Type.t * vars
   | ResInTupleList of param_id * res_in_tuple_list
   | ResInparam of param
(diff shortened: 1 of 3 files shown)


jongleb:
@ygrek  I'm sure there is no other way to solve this problem since there is no isomorphism between vars and res_expr, and I'm going to merge this pr, but in any case, please write a comment, I will address them retroactively

PR #176, opened 2025-04-10
support default value as expression

tatchi:
The [first commit](https://github.com/ygrek/sqlgg/commit/283d56f576c2b1967b6f6daa9b24f303bc98c47e) adds support for default value as expression

```sql
CREATE TABLE IF NOT EXISTS registration_feedbacks (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `user_message` TEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL DEFAULT(''),
   PRIMARY KEY (`id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
```

It currently fails with the following error

```
dune runtest
./run_test                            
Running regression tests
==> CREATE TABLE IF NOT EXISTS registration_feedbacks (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `user_message` TEXT CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL DEFAULT(''),
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
Position 3:81 Tokens: (''),
  PRIMARY KEY (`id`)
) ENGI
Error: Sqlgg.Sql_parser.MenhirBasics.Error
Errors encountered, no code generated
FAILED test/default_exression.sql
make: *** [test] Error 2
```

The [second commit](https://github.com/ygrek/sqlgg/commit/7f3e7c47fcef95f035c64680ac440f378aea20e4) add support for DEFAULT in choice expression

```sql
INSERT INTO `registration_feedbacks`
SET
  `user_message` = @user_message,
  `grant_types` = @grant_types { None { DEFAULT } | Some { @grant_types } };
```

It currently fails with the following error

```
./run_test                           
Running regression tests
==> INSERT INTO `registration_feedbacks`
SET
  `user_message` = @user_message,
  `grant_types` = @grant_types { None { DEFAULT } | Some { @grant_types } }
Position 4:49 Tokens: } | Some { @grant_types } }
Error: Sqlgg.Sql_parser.MenhirBasics.Error
Errors encountered, no code generated
FAILED test/default_exression.sql
make: *** [test] Error 2
```

### DIFF
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -391,6 +391,7 @@ default_value: e=single_literal_value
              | LPAREN e=expr RPAREN { e }
 
 set_column: name=attr_name EQUAL e=expr { name,e }
+           | name=attr_name EQUAL DEFAULT { name, Value (depends Any) }
 
 anyall: ANY | ALL | SOME { }
 
@@ -429,6 +430,7 @@ expr:
     | VALUES LPAREN n=IDENT RPAREN { Inserted n }
     | v=literal_value | v=datetime_value { v }
     | v=interval_unit { v }
+    | DEFAULT { Value (depends Any) }
     | e1=expr mnot(IN) l=sequence(expr) { poly (depends Bool) (e1::l) }
     | e1=expr mnot(IN) LPAREN select=select_stmt RPAREN { poly (depends Bool) [e1; SelectExpr (select, `AsValue)] }
     | e1=expr IN table=table_name { Tables.check table; e1 }
@@ -511,7 +513,6 @@ case_branch: WHEN e1=expr THEN e2=expr { [e1;e2] }
 like: LIKE | LIKE_OP { }
 
 choice_body: c1=LCURLY e=expr c2=RCURLY { (c1,Some e,c2) }
-          | c1=LCURLY DEFAULT c2=RCURLY { (c1,Some (Value (depends Any)),c2) }
 choice: parser_state_normal label=IDENT? e=choice_body? { let (c1,e,c2) = Option.default (0,None,0) e in ({ label; pos = (c1+1,c2) },e) }
 choices: separated_nonempty_list(pair(parser_state_ident,NUM_BIT_OR),choice) { $1 }
 
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -390,8 +390,11 @@ default_value: e=single_literal_value
              | e=datetime_value { e } (* sub expr ? *)
              | LPAREN e=expr RPAREN { e }
 
-set_column: name=attr_name EQUAL e=expr { name,e }
-          | name=attr_name EQUAL DEFAULT { name, Value (depends Any) }
+set_column: name=attr_name EQUAL e=set_column_expr { name,e }
+
+set_column_expr:
+  | e=expr { e }
+  | DEFAULT { Value (depends Any) }
 
 anyall: ANY | ALL | SOME { }
 
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -433,7 +433,6 @@ expr:
     | VALUES LPAREN n=IDENT RPAREN { Inserted n }
     | v=literal_value | v=datetime_value { v }
     | v=interval_unit { v }
-    | DEFAULT { Value (depends Any) }
     | e1=expr mnot(IN) l=sequence(expr) { poly (depends Bool) (e1::l) }
     | e1=expr mnot(IN) LPAREN select=select_stmt RPAREN { poly (depends Bool) [e1; SelectExpr (select, `AsValue)] }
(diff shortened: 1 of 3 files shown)


lib/sql_parser.mly
```diff
@@ -509,6 +511,7 @@ case_branch: WHEN e1=expr THEN e2=expr { [e1;e2] }
 like: LIKE | LIKE_OP { }
 
 choice_body: c1=LCURLY e=expr c2=RCURLY { (c1,Some e,c2) }
+          | c1=LCURLY DEFAULT c2=RCURLY { (c1,Some (Value (depends Any)),c2) }
```
jongleb, line 514: No, it shouldn't be a specific syntax, `DEFAULT` 
since an expression is supported inside `Choices`, then `DEFAULT` should be an expression in this case

jongleb: I mean I still can't just write

```
INSERT INTO `registration_feedbacks`
SET
  `user_message` = DEFAULT;
```

When you can write it , you will run the test with the usage of it inside choices and see that it also works as the result

tatchi: Thanks for the review 🙏 Is it better [`74d3cb2` (#176)](https://github.com/ygrek/sqlgg/pull/176/commits/74d3cb22cc084a8fc1af55727db94e2fd5cc9695) ?

lib/sql_parser.mly
```diff
@@ -391,6 +391,7 @@ default_value: e=single_literal_value
              | LPAREN e=expr RPAREN { e }
 
 set_column: name=attr_name EQUAL e=expr { name,e }
+           | name=attr_name EQUAL DEFAULT { name, Value (depends Any) }
```
jongleb: Almost,

```
set_column: name=attr_name EQUAL e=set_column_expr { name,e }

set_column_expr:
  | e=expr { e }
  | DEFAULT { Value (depends Any) }
```

This way would be even better.

So merge and don't forget please to squash

lib/sql_parser.mly
```diff
@@ -427,6 +430,7 @@ expr:
     | VALUES LPAREN n=IDENT RPAREN { Inserted n }
     | v=literal_value | v=datetime_value { v }
     | v=interval_unit { v }
+    | DEFAULT { Value (depends Any) }
```
jongleb, line 433: Now it is possible to write:

```
SELECT * FROM registration_feedbacks
WHERE user_message = DEFAULT;
```

That is invalid. I guess you shouldn't add it here, but then it will break `choices`, let me think about how to do better

lib/sql_parser.mly
tatchi: Thanks, done in [`439fc18` (#176)](https://github.com/ygrek/sqlgg/pull/176/commits/439fc18e5fed7323968d4183653d4713b8843c2b)

lib/sql_parser.mly
tatchi: Good point. [`76decd8` (#176)](https://github.com/ygrek/sqlgg/pull/176/commits/76decd846f2c45a404803d75251c17243ed8ff16) disallow this syntax (parsing fails) while not breaking any current test. I hope that's fine 😅

jongleb: Now you can write 
```
SELECT ....
FROM ....
WHERE ... @param Case1{...} | Case2 { DEFAULT }
```

jongleb: Perhaps it is worthwhile to build a separate pass for set then, but not now,  for the first step better limit yourself to SET for only the standard syntax

jongleb: Let's remove `choice_body` for now and I'll merge it

jongleb:
https://github.com/ygrek/sqlgg/blob/master/lib/sql.ml#L501
https://github.com/ygrek/sqlgg/blob/master/lib/sql_parser.mly#L493
You can do the similar way but you also need a context if it's set or not


Or 

```


%inline choice_structure(E):
  parser_state_normal label=IDENT? e=choice_body_structure(E)? 
  { let (c1,e,c2) = Option.default (0,None,0) e in ({ label; pos = (c1+1,c2) }, e) }

%inline choice_body_structure(E):
  c1=LCURLY e=E c2=RCURLY { (c1, Some e, c2) }
```

Something like this to make choices reusable , but then you will always have to adapt all the features to this default, which is of course a bad idea for the thing of default, so we will choose the 1st option

Khady:
status on this one?

jongleb:
@Khady implemented here https://github.com/ygrek/sqlgg/pull/189 , @tatchi so I am closing it

jongleb:
https://github.com/ygrek/sqlgg/pull/189

PR #205, opened 2025-06-02
Fixes of Support different dialects

jongleb:
### Description
This PR adds fixes of [Support different dialects](https://github.com/ygrek/sqlgg/pull/204)


lib/dialect.ml
```diff
@@ -6,6 +6,18 @@ type feature =
   | CreateTableAsSelect
 [@@deriving show { with_path = false }]
 
+let feature_to_string = function
+  | Collation -> "collation"
+  | JoinOnSubquery -> "join_on_subquery"
+  | CreateTableAsSelect -> "create_table_as_select"
```
ygrek, line 12: use [@name] to make deriving generate this mapping

PR #208, opened 2025-07-01
Support Json type

jongleb:
# JSON Support for sqlgg

This pull request adds comprehensive JSON support to sqlgg - a typed SQL interface generator for OCaml.

### 1. New Data Types
- **`Json`** - for storing JSON data
- **`Json_path`** - for JSON paths (e.g., `$.user.name`)  
- **`One_or_all`** - for search parameters (`'one'` or `'all'`)

### 2. JSON Path Parser
Added a complete JSON path parser (`json_path/`) supporting:
- Basic paths: `$.name`, `$[0]`, `$[*]`
- Complex expressions: `$.users[*].email`, `$[1 to 3]`
- Special operators: `last`, `**` (recursive search)
- Combinators for programmatic path construction

### 3. MySQL JSON Functions
Implemented MySQL JSON functions with proper typing:

```sql
-- Data extraction
SELECT JSON_EXTRACT(data, '$.name') FROM users;

-- JSON modification
UPDATE users SET data = JSON_SET(data, '$.active', true);

-- JSON search
SELECT JSON_SEARCH(data, 'one', 'pattern%');

-- JSON creation
SELECT JSON_OBJECT('name', 'Alice', 'age', 30);
```

### 4. Enhanced Type System
- **Static validation**: JSON literals are validated at compile time
- **Variadic functions with grouped arguments**: Support for pattern `JSON_SET(json, path, value, path, value, ...)` where arguments come in fixed groups rather than simple repetition like `CONCAT(a, b, c, ...)`
- **Nullable semantics**: Distinction between SQL NULL and JSON null

#### Type Safety
```ocaml
(* JSON correctness is statically verified *)
JSON_SET(data, '$.name', "valid")  (* ✓ *)
JSON_SET(data, 'invalid path', x)  (* ✗ compile error *)
```

#### Path Combinators
```ocaml
let open Sqlgg_json_path.Ast.Syntax in
let path = root / ~."users" / any / ~."settings" in
(* Equivalent to "$.users[*].settings" *)
```

#### Meta Attributes
Support for special DDL annotations:
```sql
-- [sqlgg] json_null_kind=false
data JSON NOT NULL
```
This is a way to specify that your json will not include null kind. not to be confused with nullability of the column itself. Since in sql itself there is no way to specify that json kind null will be included or not, this flag is added

### DIFF
--- a/test/cram/json_functions.compare.ml
+++ b/test/cram/json_functions.compare.ml
@@ -1,0 +1,198 @@
+module Sqlgg (T : Sqlgg_traits.M) = struct
+
+  module IO = Sqlgg_io.Blocking
+
+  let create_people db  =
+    T.execute db ("CREATE TABLE people (\n\
+  id INT AUTO_INCREMENT PRIMARY KEY,\n\
+  data JSON\n\
+)") T.no_params
+
+  let create_people_data_json_kind_never_null_but_col_nullable db  =
+    T.execute db ("CREATE TABLE people_data_json_kind_never_null_but_col_nullable (\n\
+  id INT AUTO_INCREMENT PRIMARY KEY,\n\
+  data JSON\n\
+)") T.no_params
+
+  let create_people_data_json_kind_never_null_and_col_strict db  =
+    T.execute db ("CREATE TABLE people_data_json_kind_never_null_and_col_strict (\n\
+  id INT AUTO_INCREMENT PRIMARY KEY,\n\
+  data JSON NOT NULL\n\
+)") T.no_params
+
+  let one db  =
+    T.execute db ("INSERT INTO people (data) VALUES\n\
+  (JSON_OBJECT(\n\
+    'name', 'Alice',\n\
+    'age', 30,\n\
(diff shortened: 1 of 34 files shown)


test/cram/json_functions.compare.ml
```diff
@@ -0,0 +1,198 @@
...
+      let r_acc = ref acc in
+      IO.(>>=) (T.select db ("SELECT\n\
+  id,\n\
+  JSON_SET(data, '$.active', TRUE) AS data_with_active\n\
+FROM people_data_json_kind_never_null_but_col_nullable") T.no_params (fun x -> r_acc := invoke_callback x !r_acc))
+      (fun () -> IO.return !r_acc)
+
+    let five db  callback acc =
+      let invoke_callback stmt =
+        callback
```
jongleb, line 132: Most of the PR is taken up by tests, because at first glance there is not much code, but new types have been added

jongleb:
@ygrek 
I haven't merged this into master, but moved it to a separate branch as we see. I'm looking forward to your review. Please take a look when you have time.

PR #221, opened 2025-08-12
Add constraint-based cardinality detection

s-kybound:
Following example is in `Sqlite3` syntax, but the behaviour is generic across all dialects:

```sql
create table foo (
    id int primary key,
    name text
);

-- @get_name_by_id
select name from foo where id = @id;
```

`sqlgg` generated code is as such:
```ocaml
let get_name_by_id db ~id callback =
    let invoke_callback stmt =
      callback
        ~name:(T.get_column_Text stmt 0)
    in
    let set_params stmt =
      T.set_param_Int p id;
      T.finish_params p
    in
    T.select db ("select name from foo where id = @id") set_params invoke_callback
```
Since `id` is a primary key, by comparing equality of `id` we should expect a maximum of one value returned. As such it would be nice to have the ocaml code generated be
```ocaml
  let get_name_by_id db ~id =
    let get_row stmt =
      (T.get_column_Text stmt 0)
    in
    let set_params stmt =
      T.set_param_Int p id;
      T.finish_params p
    in
    T.select_one_maybe db ("select name from foo where id = @id limit 1") set_params get_row
```
which avoids the callback function.

This PR uses constraints in defined in the schema to determine a more accurate cardinality for SELECT statements that use comparisons based on values with UNIQUE or PRIMARY KEY constraints.


lib/syntax.ml
```diff
@@ -800,6 +801,74 @@ and eval_select env { columns; from; where; group; having; } =
...
+            let columns_in_eql_check =
+            match v1, v2 with
+            | Column _, Column _ -> [] (* Columns may refer to each other in foreign
+                                          key constraints, so don't add any of them for now. 
+                                          TODO: consider foreign key constraints *)
+            | Column col1, _ -> [col1]
+            | _ , Column col2 -> [col2]
+            | _ -> []
+            in
+            aux (columns_in_eql_check @ acc) expr_list
```
jongleb, line 821:
```ocaml
let rec aux = function
        | Fun { kind = Comparison Comp_equal; parameters = [v1; v2]; _ } ->
            (match v1, v2 with
            | Column _, Column _ -> []
            | Column c, _ | _, Column c -> [c]
            | _ -> [])
        | Fun { kind = Logical And; parameters = [x; y]; _ } -> aux x @ aux y
        | _ -> []
      in
      aux expr
```
Why not collect columns directly?

lib/syntax.ml
```diff
@@ -800,6 +801,74 @@ and eval_select env { columns; from; where; group; having; } =
...
+          | Fun { kind = Logical And; parameters = [x; y]; _ } -> aux acc (x :: y :: expr_list)
+            (* as OR, XOR can easily propagate to tautology, we avoid checking them *)
+          | Fun { kind = Logical Or; _ } -> aux acc expr_list
+          | Fun { kind = Logical Xor; _ } -> aux acc expr_list
+          | _ -> aux acc expr_list
```
jongleb, line 826: Are these Or and Xor cases left here as a future TODO?
otherwise it seems that a wildcard is enough, if they left it on purpose then let them be

lib/syntax.ml
s-kybound: I wanted to keep the auxillary function tail-recursive :sweat_smile:

lib/syntax.ml
```diff
@@ -800,6 +801,74 @@ and eval_select env { columns; from; where; group; having; } =
...
+      in
+      let composite_constraint_sets = List.concat_map get_composite_sets_over_column column_constraints in
+      match composite_constraint_sets with
+      | [] -> false
+      | _ -> 
+      List.exists (fun composite_set -> 
+        List.for_all 
+        (fun composite_component -> List.mem composite_component ids) 
+        composite_set
+      ) composite_constraint_sets
```
jongleb, line 860:
```ocaml
let satisfies_composite_constraint () =
      let open Constraint.StringSet in
      let ids_set = of_list (List.map (fun col -> col.cname) columns_in_where) in
      List.exists
        (fun cs ->
          List.exists
            (function
              | Constraint.Composite (CompositePrimary s)
              | Constraint.Composite (CompositeUnique s) -> subset s ids_set
              | _ -> false)
            (Constraints.elements cs)
        )
        column_constraints
```
It seems that you can do without `concat_map`, but you can leave your own version

lib/syntax.ml
jongleb: I see, then leave yours

jongleb (approved):
Nice job!

PR #225, opened 2025-08-15
Hotfix: consider only relevant constraints

s-kybound:
The previous PR would match ANY constraint matched in an SQL query, even if the matched constraint does not correlate to the current table being selected.

### DIFF
--- a/lib/syntax.ml
+++ b/lib/syntax.ml
@@ -856,19 +856,17 @@ and eval_select env { columns; from; where; group; having; } =
     satisfies_single_value_constraint () || satisfies_composite_constraint ()
   in
   let cardinality =
-    match from with
-    | None -> (if where = None then `One else `Zero_one)
-    | Some ((from, _), _) ->
-    match group = [] && exists_grouping columns && not @@ exists_windowing columns with
-    | true -> `One
-    | false -> 
-    match from with
-    | `Table t -> begin
-      match where with
-      | None -> `Nat
-      | Some where -> if satisfies_some_relevant_constraint t where env then `Zero_one else `Nat 
-    end
-    | _ -> `Nat
+    match from, where with
+    | None, None ->
+      `One
+    | None, Some _ ->
+      `Zero_one
+    | Some _, _ when group = [] && exists_grouping columns && not (exists_windowing columns) ->
+      `One
+    | Some ((`Table t, _), _), Some w when satisfies_some_relevant_constraint t w env ->
+      `Zero_one
+    | Some _, _ ->
+      `Nat
   in
   let p4 = get_params_l env group in
   let p5 = get_params_opt env having in


lib/syntax.ml
```diff
@@ -855,12 +856,19 @@ and eval_select env { columns; from; where; group; having; } =
...
+    match group = [] && exists_grouping columns && not @@ exists_windowing columns with
+    | true -> `One
+    | false -> 
+    match from with
+    | `Table t -> begin
+      match where with
+      | None -> `Nat
+      | Some where -> if satisfies_some_relevant_constraint t where env then `Zero_one else `Nat 
+    end
+    | _ -> `Nat
```
jongleb, line 871:
```ocaml
  let cardinality =
    match from, where with
    | None, None ->
      `One
    | None, Some _ ->
      `Zero_one
    | Some _, _ when group = [] && exists_grouping columns && not (exists_windowing columns) ->
      `One
    | Some ((`Table _, _), _), None ->
      `Nat
    | Some ((`Table t, _), _), Some w
      when satisfies_some_relevant_constraint t w env ->
      `Zero_one
    | Some ((_, _), _), (None | Some _) ->
      `Nat
  in
```
fix me if I got it wrong, but I want to do it in the most "table" way possible

lib/syntax.ml
```diff
@@ -830,6 +830,7 @@ and eval_select env { columns; from; where; group; having; } =
     (* identify if the set of columns used in WHERE clause contains/represents constraints *)
     let columns_in_where = get_all_eql_checks where in
     let attributes_in_where = List.map (resolve_column ~env) columns_in_where in
+    let attributes_in_where = List.filter (fun attr -> List.mem table attr.Schema.Source.Attr.sources) attributes_in_where in
     let column_constraints = 
       List.map (fun (attr : table_name Source.Attr.t) -> attr.attr.extra) attributes_in_where 
     in
```
jongleb, line 836:
```ocaml
let column_constraints =
      get_all_eql_checks where
      |> List.to_seq
      |> Seq.map (resolve_column ~env)
      |> Seq.filter (fun attr ->
          List.mem table attr.Schema.Source.Attr.sources)
      |> Seq.map (fun (attr : table_name Source.Attr.t) ->
          attr.attr.extra)
      |> List.of_seq
      in
```
It seems to me that here we need to express several iterations of the list using seq and get rid of the intermediate variables, the second one is probably purely stylistic.

lib/syntax.ml
s-kybound:
```ocaml
| Some ((`Table _, _), _), None ->
      `Nat
```
we can get rid of this check, otherwise looks good to me! will add and verify.

jongleb (approved):

PR #270, opened 2026-02-25
Dynamic select in regular select syntax

jongleb:
### Description

This PR removes the old approach to Dynamic Select. Instead, there are no choice syntaxes. Only standard sql select.


doc/sql/dynamic-select.md
```diff
@@ -75,188 +84,321 @@ let all_fields =
...
+```ocaml
+open List_products_col
+
+let combined =
+  let+ i = id
+  and+ n = name
+  and+ p = price in
+  (i, n, p)
+
+let () = Db.list_products conn ~col:combined ~min_stock:5L (fun ~col ->
```
ygrek, line 118: ~col is kind of strange name (~select better?) and also do we really need a named param here, it is impossible to mix it up with anything because of unique type. Same applies to the callback parameter

doc/sql/dynamic-select.md
```diff
@@ -12,52 +12,61 @@ Dynamic Select is currently supported **only in OCaml** code generation (`-gen c
 
 ## Overview
 
-Dynamic Select allows you to choose which columns to SELECT at runtime while maintaining full type safety. Unlike regular `@choice` which selects between SQL fragments, Dynamic Select generates a module with composable field selectors that can be combined using applicative combinators.
+Dynamic Select allows you to choose which columns to SELECT at runtime while maintaining full type safety. Just write a regular `SELECT` query and add the `dynamic_select=true` metadata flag — every column in the select list becomes a composable field that can be picked or combined at runtime using applicative combinators.
```
ygrek, line 15: should describe more why it is more useful than just using subset of columns in callback (because it will eliminate unused parts of the query and send smaller query to the server)

doc/sql/dynamic-select.md
```diff
@@ -12,52 +12,61 @@ Dynamic Select is currently supported **only in OCaml** code generation (`-gen c
...
-SELECT id, @col { Name { name } | Price { price } | Category { category } }
-FROM products WHERE id = @id;
+SELECT id, name, price, category FROM products WHERE id = @id;
 ```
 
-This generates a module `Select_product_col` with field constructors:
+This generates a module `Select_product_col` with a field for each column:
 
 ```ocaml
 module Select_product_col : sig
```
ygrek, line 30: it would be easier to use if the query itself was part of this module, then no need to open extra module. like: `Db.Select_product.(let fields = let+ a .. and+ ... in (a,b) in select db fields (fun (a,b) -> ..))` wdyt

ygrek: or even `Db.Select_product.(select db (let+ a and+ b in a,b) (fun (a,b) -> ...))`

doc/sql/dynamic-select.md
```diff
@@ -75,188 +84,321 @@ let all_fields =
...
+SELECT *, id + 2 AS id_plus FROM products WHERE id = @id;
+```
+
+## Module-Wrapped Columns
+
+Works with `[sqlgg] module=` annotations on columns:
 
 ```sql
+CREATE TABLE products_wrapped (
+    -- [sqlgg] module=Product_id
```
ygrek, line 347: we need some shortcut form for this annotation `-- @module=Product_id`?

doc/sql/dynamic-select.md
ygrek: but actually i understand it will not currently eliminate unused joins and i am not sure if it is even realistic to encode in a dynamic setting (but can straightforwardly do at genereation-time)

PR #275, opened 2026-05-01
Fix missing version in pin-depends

Niols:
Without it, all `opam` commands print:

```
error  3: File format error in 'pin-depends' at line 44, column 4: while expecting versioned package:
          OpamPackage.of_string
```

### DIFF
--- a/sqlgg.opam
+++ b/sqlgg.opam
@@ -41,7 +41,7 @@ conflicts: [
 ]
 
 pin-depends: [
-  [ "mariadb"
+  [ "mariadb.dev"
     "git+https://github.com/ocaml-community/ocaml-mariadb.git#e9f9ac750abd5adebb589da5e553e5c45d8ec066"
   ]
 ]


ygrek:
@jongleb is there a released version of mariadb that we can depend upon?

Niols:
I checked and `sqlgg` as of latest master doesn't compile against `mariadb.1.3.0`, [the latest version of the OCaml MariaDB bindings](https://github.com/ocaml-community/ocaml-mariadb/releases/tag/1.3.0). That being said, that version is from May 2025, so I'm not sure there are releases anymore.

The error:
```
File "impl/ocaml/mariadb/sqlgg_mariadb.ml", line 169, characters 12-20:
169 |           | `Int64 x -> Int64.to_string x
                  ^^^^^^^^
Error: This pattern matches values of type [? `Int64 of 'a ]
       but a pattern was expected which matches values of type value
       The second variant type does not allow tag(s) `Int64
```

Niols:
Rebased on most recent `master`.

ygrek:
thanks

PR #276, opened 2026-05-01
Add support for PostgreSQL-style `ALTER COLUMN` syntax

Niols:
PostgreSQL (and standard SQL) uses `ALTER COLUMN` instead of MySQL's `CHANGE COLUMN` / `MODIFY COLUMN` for modifying existing columns. This adds parsing and schema evaluation for the five main sub-commands:

- `ALTER COLUMN col TYPE newtype`
- `ALTER COLUMN col SET NOT NULL`
- `ALTER COLUMN col DROP NOT NULL`
- `ALTER COLUMN col SET DEFAULT expr`
- `ALTER COLUMN col DROP DEFAULT`

These can be combined with other actions in a single `ALTER TABLE` statement, eg.:

```
ALTER TABLE t
  DROP COLUMN old_col,
  ALTER COLUMN col TYPE SMALLINT,
  ALTER COLUMN col SET NOT NULL;
```

Migration generation (inverse actions + SQL fragments) is also supported for all five variants.

---

I hope everything looks well; please let me know if you want me to change anything!

### DIFF
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -369,6 +369,19 @@ alter_action: ADD COLUMN? col=maybe_parenth(column_def) pos=alter_pos { `Add (co
             | DROP CHECK name=IDENT { `DropConstraint name }
             | CHANGE COLUMN? old_name=IDENT column=column_def pos=alter_pos { `Change (old_name,column,pos) }
             | MODIFY COLUMN? column=column_def pos=alter_pos { `Change (column.Alter_action_attr.name,column,pos) }
+            (* TYPE is not a reserved word (columns may be named `type`), so match it as IDENT *)
+            | ALTER COLUMN? col=IDENT kw=IDENT t=located_sql_type
+                { if String.lowercase_ascii kw <> "type" then failwith ("expected TYPE after ALTER COLUMN, got " ^ kw);
+                  `AlterColumnPG (col, { Alter_column_pg.typ = Some t; not_null = None; default = None }) }
+            | ALTER COLUMN? col=IDENT SET NOT NULL
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = Some true; default = None }) }
+            | ALTER COLUMN? col=IDENT DROP NOT NULL
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = Some false; default = None }) }
+            | ALTER COLUMN? col=IDENT SET DEFAULT e=default_value
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = None;
+                    default = Some (Some (make_located ~value:e ~pos:($startofs, $endofs))) }) }
+            | ALTER COLUMN? col=IDENT DROP DEFAULT
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = None; default = Some None }) }
             | opts=ttl_option+ { `TtlOptions (opts, ($startofs, $endofs)) }
             | REMOVE TTL { `RemoveTtl ($startofs, $endofs) }
             | either(DEFAULT,pair(CONVERT,TO))? cs=charset c=collate? { `Default_or_convert_to (cs, c) }
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -356,14 +356,13 @@ maybe_as_with_detupled: AS? name=IDENT names=sequence(IDENT)? { name, names }
 maybe_parenth(X): x=X | LPAREN x=X RPAREN { x }
 
 alter_column_pg_spec:
-  (* TYPE is not a reserved word (columns may be named `type`), so match it as IDENT *)
   | kw=IDENT t=located_sql_type
       { if String.lowercase_ascii kw <> "type" then failwith ("expected TYPE after ALTER COLUMN, got " ^ kw);
-        { Alter_column_pg.typ = Some t; not_null = None; default = None } }
-  | SET NOT NULL { { Alter_column_pg.typ = None; not_null = Some true; default = None } }
-  | DROP NOT NULL { { Alter_column_pg.typ = None; not_null = Some false; default = None } }
-  | SET DEFAULT e=default_value { { Alter_column_pg.typ = None; not_null = None; default = Some (Some (make_located ~value:e ~pos:($startofs, $endofs))) } }
-  | DROP DEFAULT { { Alter_column_pg.typ = None; not_null = None; default = Some None } }
+        Alter_column_pg.Set_type t }
+  | SET NOT NULL { Alter_column_pg.Set_not_null }
+  | DROP NOT NULL { Alter_column_pg.Drop_not_null }
+  | SET DEFAULT default_value { Alter_column_pg.Set_default }
+  | DROP DEFAULT { Alter_column_pg.Drop_default }
 
 alter_action: ADD COLUMN? col=maybe_parenth(column_def) pos=alter_pos { `Add (col,pos) }
             | ADD PRIMARY KEY cols=sequence(IDENT) { `AddPrimaryKey cols }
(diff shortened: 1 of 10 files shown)


ygrek:
looks very repetitve, i believe all of these + the existing ones can be expressed as a single variant (e.g. can take old type, apply modifications, then use `Change` ctor)

Niols:
Yes, I was wondering about that. It felt a bit wrong to encode granular instructions into something that just updates everything at once, but since CHANGE and MODIFY are already here, then why not. I was also worried about code generation, but in retrospect I think that it isn't really a problem.

Let me have a look and propose an updated version in the coming days!

Niols:
I thought about this some more, and I don't really see how that would be doable, at least not just as taking the old type, applying modifications, then using the `Change` constructor, because taking the old type isn't really possible in the parser anyway. What I could imagine is to introduce extend the `Change` constructor such that it can encode some or all of the changes independently. MySQL's `MODIFY` would be everything, MySQL's `CHANGE` everything but the name, and PostgreSQL's `ALTER COLUMN` would only affect one of the coordinates. I'm not sure that would play well with everything else because I don't understand much, but does that sound like an approach worth pursuing?

ygrek:
my thinking is as follows: in the current PR it is more like CST (variants directly correspond to syntax), which can be useful but currently all the other parsing outputs more abstract nodes, and that would help to have less changes in tables.ml. But then I realize in Gen_migrations.sql_to_fragment we actually seem to need to preserve exact sql syntax used... Probably best option would be to support dialect flag when generating migrations and convert from AST to the specific dialect, but this increases a scope of this PR quite a bit.
So I propose you keep one separate ctor for the pgsql alter but the actual alter change is described with one record ie smth like `AlterPQ of { nullability: bool option; typ : typ option; etc }`

Niols:
That sounds like a fair compromise. Indeed, I also would see something like what you describe, with an AST that is fairly generic and then a dialect flag to know how to generate SQL back.

The PG-specific constructor that you propose could serve as a base for this future, generic mechanism, while keeping the current change small enough. I will make an update in those lines.

jongleb:
Sorry, I didn't see this PR right away. And maybe I'm intruding, but 
lib/dialect.ml already has a dialect identifier based on syntax. Just as it's used for the standard generator in subsequent steps, it can also be used for generating migrations

Niols:
What do you think of https://github.com/ygrek/sqlgg/pull/276/commits/f13bbfa8ca90565d55979b51cb6a3d8d084f2787?

A few notes:
- I have merged the `` `AlterColumn* `` variants but not the `alter_column_*` helpers in `lib/tables.ml`. I think the helpers make more sense as small units, even if we are free.
- I haven't tried generating migrations while using the dialect and I have kept `` `AlterColumnPG `` separate from the existing `` `Change `` constructor. I don't think it would be that much work to merge the two, but I wanted to keep the PR focused.
- I am not really familiar with the code base, and there are some things that I did that look fishy to me. I will open a review and point at those things.

Let me know if you would like me to change anything. I could also attempt merging `` `Change `` and `` `AlterColumnPG `` (and `` `RenameColumn `` presumably!) into the same `` `AlterColumn `` constructor if you thought that that was useful in this PR.

src/gen_migrations.ml
```diff
@@ -156,6 +177,21 @@ let action_to_sql_fragment (action : Sql.alter_action) =
...
+    let parts = List.filter_map Fun.id [
+      Option.map (fun t ->
+        sprintf "ALTER COLUMN %s TYPE %s" (quote_id col_name) (source_type_kind_to_sql t.Sql.value.Sql.collated)
+      ) changes.Sql.Alter_column_pg.typ;
+      Option.map (fun b ->
+        if b then sprintf "ALTER COLUMN %s SET NOT NULL" (quote_id col_name)
+        else sprintf "ALTER COLUMN %s DROP NOT NULL" (quote_id col_name)
+      ) changes.not_null;
+      Option.map (function
+        | Some _ -> sprintf "ALTER COLUMN %s SET DEFAULT (* unknown *)" (quote_id col_name)
```
Niols: Not sure about this `(* unknown *)` since that produces invalid code. It was inspired by the `(* unsupported: unknown previous charset *)` above, but at least that one replaced a whole statement? Do you prefer that we replace a whole statement here too?

src/gen_migrations.ml
```diff
@@ -111,6 +111,27 @@ let inverse_action table_name (columns : Tables.column list) (action : Sql.alter
...
+    let inv_typ = Option.map (fun _ ->
+      let old = Sql.Alter_action_attr.from_attr entry.attr |> enrich_with_source_kind entry.source_kind in
+      match old.Sql.Alter_action_attr.kind with
+      | Some k -> k
+      | None -> Sql.make_located ~pos:(0,0)
+          ~value:(Sql.make_collated ~collated:(Sql.Source_type.Infer entry.attr.domain.Sql.Type.t) ())
+    ) changes.Sql.Alter_column_pg.typ in
+    let inv_not_null = Option.map (fun _ ->
+      Sql.Type.is_strict entry.attr.domain
+    ) changes.not_null in
```
Niols: This is here trying to respect that the inverse of `SET NOT NULL` is only `DROP NOT NULL` if the column was nullable before, otherwise it's a no-op, but I'm not sure if that's considerations that we should worry about?

lib/tables.ml
```diff
@@ -124,6 +124,52 @@ let add_primary_key name ~cols:col_names =
...
+      if c.attr.Sql.name = col_name then
+        { c with attr = { c.attr with extra = Sql.Constraints.add WithDefault c.attr.extra } }
+      else c
+    ) cols)
+
+let alter_column_drop_default name ~col_name =
+  alter name (fun cols ->
+    List.map (fun c ->
+      if c.attr.Sql.name = col_name then
+        { c with attr = { c.attr with extra = Sql.Constraints.remove WithDefault c.attr.extra } }
```
Niols: I am tempted to introduce a helper to update `attr` in depth, since the pattern shows up everywhere. Let me know if you want this.

Niols:
Rebased on most recent `master` and resolved conflicts. Is there anything to change to this PR?

lib/sql_parser.mly
```diff
@@ -367,6 +367,17 @@ alter_action: ADD COLUMN? col=maybe_parenth(column_def) pos=alter_pos { `Add (co
...
+                { `AlterColumnPG (col, { Alter_column_pg.typ = Some t; not_null = None; default = None }) }
+            | ALTER COLUMN? col=IDENT SET NOT NULL
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = Some true; default = None }) }
+            | ALTER COLUMN? col=IDENT DROP NOT NULL
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = Some false; default = None }) }
+            | ALTER COLUMN? col=IDENT SET DEFAULT e=default_value
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = None;
+                    default = Some (Some (make_located ~value:e ~pos:($startofs, $endofs))) }) }
+            | ALTER COLUMN? col=IDENT DROP DEFAULT
+                { `AlterColumnPG (col, { Alter_column_pg.typ = None; not_null = None; default = Some None }) }
```
jongleb: They have the same tail, right?

```
| ALTER COLUMN? col=IDENT spec=alter_column_pg_spec { `AlterColumnPG (col, spec) }
```

and then

```
alter_column_pg_spec:
    TYPE t=located_sql_type
      { { Alter_column_pg.typ = Some t; not_null = None; default = None } }
  | SET NOT NULL
      { { Alter_column_pg.typ = None; not_null = Some true; default = None } }
  | DROP NOT NULL
      { { Alter_column_pg.typ = None; not_null = Some false; default = None } }
  | SET DEFAULT e=default_value
      { { Alter_column_pg.typ = None; not_null = None;
          default = Some (Some (make_located ~value:e ~pos:($startofs(e), $endofs(e)))) } }
  | DROP DEFAULT
      { { Alter_column_pg.typ = None; not_null = None; default = Some None } }
```

Niols: Nice indeed! Done in 9a9bb8195498c746f3207dd238b0b5624c13a372.

jongleb:
I ventured to remove the record after all while rebasing onto master. I brought back the previous version with variant type because `tables.ml` was changed a lot in the auto-migrations [PR](https://github.com/ygrek/sqlgg/pull/282). New functions were added there, and I reused them. So, it looks cleaner without the record

jongleb:
Thank you , I am merging

Niols:
Thank you!

PR #279, opened 2026-05-11
Gen migrate add varchar(N), blob(N) default, collates

jongleb:
### Description

This PR adds proper support for VARCHAR(N), CHAR(N), VARCHAR2, BLOB(N), VARBINARY(N), TINYTEXT/MEDIUMTEXT/LONGTEXT (and their blob variants), and DECIMAL(p) in gen_migrations.

Previously these types were normalized into plain TEXT/BLOB/DECIMAL, so inverse migrations lost the original type flavor and length information.

Inverse migrations now also preserve COLLATE and DEFAULT clauses when reconstructing columns from the stored schema. Before this change, those attributes were silently dropped.

For now, DEFAULT values are stored as raw SQL fragments and replayed verbatim in inverse migrations. This is temporary. Once DEFAULT expressions are typechecked against the column type, they can be represented as structured values instead of opaque SQL text.

### DIFF
--- a/lib/parser_state.ml
+++ b/lib/parser_state.ml
@@ -11,3 +11,18 @@ module Stmt_metadata = struct
   let find_all k = Hashtbl.find_all stmt_metadata k
   let reset () = Hashtbl.reset stmt_metadata
 end
+
+let current_lexbuf : Lexing.lexbuf option ref = ref None
+
+let with_lexbuf lexbuf f =
+  let saved = !current_lexbuf in
+  current_lexbuf := Some lexbuf;
+  Fun.protect f ~finally:(fun () -> current_lexbuf := saved)
+
+let extract_source (start_, end_) =
+  Stdlib.Option.bind !current_lexbuf (fun lexbuf ->
+    let len = end_ - start_ in
+    if len > 0 && start_ >= 0 && end_ <= lexbuf.Lexing.lex_buffer_len then
+      Some (Bytes.sub_string lexbuf.lex_buffer start_ len)
+    else
+      None)
--- a/lib/sql_parser.mly
+++ b/lib/sql_parser.mly
@@ -19,6 +19,9 @@
     List.filter_map param l, List.mem (`Limit,`Const 1) l
 
   let poly name ret parameters = Fun { fn_name = name; kind = (F (Typ ret, List.map (fun _ -> Var 0) parameters)); parameters; is_over_clause = false }
+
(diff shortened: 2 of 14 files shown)


lib/parser_state.ml
```diff
@@ -11,3 +11,16 @@ module Stmt_metadata = struct
...
+let with_lexbuf lexbuf f =
+  let saved = !current_lexbuf in
+  current_lexbuf := Some lexbuf;
+  Fun.protect f ~finally:(fun () -> current_lexbuf := saved)
+
+let extract_source (start_, end_) =
+  Option.map (fun lexbuf ->
+    let offset = start_ - lexbuf.Lexing.lex_abs_pos in
+    String.trim (Bytes.sub_string lexbuf.Lexing.lex_buffer offset (end_ - start_))
+  ) !current_lexbuf
```
jongleb: The issue even isn't that the default isn't unified. It's that it's any expression. This means we need `expr->string`, which fully restores the expression as an SQL string. And that requires a separate PR. So, we temporarily capture it as a string. Hence the hack in the parser state.

lib/sql_parser.mly
```diff
@@ -427,7 +439,13 @@ column_def_extra: PRIMARY? KEY { Some (Alter_action_attr.Syntax_constraint Prima
...
                 | UNIQUE KEY? { Some (Alter_action_attr.Syntax_constraint Unique) }
                 | AUTOINCREMENT { Some (Alter_action_attr.Syntax_constraint Autoincrement) }
-                | DEFAULT def=default_value { Some (Alter_action_attr.Default (make_located ~value:def ~pos:($startofs, $endofs))) }
+                | DEFAULT def=default_value {
+                    let pos = ($startofs(def), $endofs(def)) in
+                    Some (Alter_action_attr.Default {
+                      expr = make_located ~value:def ~pos;
+                      sql = Parser_state.extract_source pos;
+                    })
+                  }
```
jongleb, line 448: A temporary solution is to keep the DEFAULT expression as-is until its type is unified with the column type.

ygrek:
i think it is ok to keep column attributes as text for the purpose of migration gen (use structured repr for other uses like typechecking etc), because it could be some custom database specific attributes that we don't parse and it seems better to do something reasonable (pass them verbatim) than just fail

jongleb:
https://github.com/ygrek/sqlgg/pull/282
Closed due to a more elegant solution

PR #280, opened 2026-06-03
Always add parentheses inside Choice expressions

jongleb:
### Description

Choice branches were previously substituted into the SQL as plain text, which meant surrounding operators could affect how the branch expression was interpreted. For example:

```sql
SELECT 1 WHERE FALSE AND @b { None { TRUE } | Some { FALSE OR TRUE } };
```

was effectively parsed as:

```sql
FALSE AND FALSE OR TRUE
```

instead of:

```sql
FALSE AND (FALSE OR TRUE)
```

To avoid this, branch expressions are now wrapped in parentheses during code generation. This preserves the intended grouping and removes the need to add parentheses manually.

Since a branch body always represents a complete expression, wrapping it in parentheses does not change its meaning. The SQL contexts where parentheses would have different semantics do not apply here. Subqueries and `IN` clauses are handled separately, row constructors require commas between elements, and ordinal references in `ORDER BY` and `GROUP BY` continue to work correctly when parenthesized.

The only visible change is the presence of some redundant parentheses in generated SQL.

Dynamic select column lists and `ASC`/`DESC` direction branches are unchanged and remain unwrapped.

mysql 
https://www.db-fiddle.com/f/eep2E8Q8iygsC7iRndGLJo/0

postgresq,
https://www.db-fiddle.com/f/ei6nr6ZxVDdGe2hFQY6L2c/0

In both exmaples is clear that the brackets do not affect the `ORDER BY` in any way.


jongleb:
There are more changes in the tests than in the code. Actually, the only change is in the generator, 1 line. I'm merging, but if there's anything wrong @ygrek , please leave a comment and I'll fix it. 

Only if, of course, you agree that we always leave the parentheses and don't try to remove them for the sake of "cosmetics" of the resulting query. Like: `((TRUE OR FALSE))` to be `(TRUE OR FALSE)` when `Some { (TRUE OR FALSE) }`

PR #281, opened 2026-06-05
Add support for PostgreSQL-style `CREATE TYPE`

Niols:
This PR builds on top of https://github.com/ygrek/sqlgg/pull/276. Although they are in practice independent changes, they both aim at improving PostgreSQL support in sqlgg, and I have added the new commit on top of the previous one. When reviewing, I recommend making sure to only read the last commit, or, even better, to wait before #276 is merged before reviewing anything here.

I hope the change is welcome; please let me know if there is anything that you would want me to do differently, I will be happy to update my PR.

## Target

PostgreSQL allows declaring a named type with `CREATE TYPE`, then referring to it elsewhere. This is most useful for enums:

```sql
CREATE TYPE "color" AS ENUM ('red', 'green', 'blue');

CREATE TABLE "shirt" (
  "id" INT NOT NULL,
  "color" "color" NOT NULL
);

SELECT "id" FROM "shirt" WHERE "color" = 'red';
```

and in fact necessary, as inline `ENUM` is just not accepted. Unlike MySQL's inline `ENUM(...)` columns, this is a separate statement that introduces a name to be looked up later. It cannot be handled at the leaf of the type grammar alone: it requires a registry that one statement populates and a later statement consults.

## Change

The main change is the new module `lib/types.ml`, similar to `lib/tables.ml` (albeit much simpler) but for types. Encountering a new `CREATE TYPE` adds to a global registry found in that module; the counterpart `DROP TYPE` removes from it, and we look types up during analysis of `CREATE TABLE`. The rest of the change is what one expects, namely adding those syntaxes to the grammar and following the new types everywhere. I add some Cram-style tests for the feature in `test/cram/create_type.t`.

## Limitations

- `CreateType` / `DropType` carry no source position, so the dialect-mismatch error reads `... at` with an empty tail.
- `DROP TYPE IF EXISTS` parses, but the flag is discarded, so it still errors on missing names. This matches the pre-existing behaviour of `DROP TABLE IF EXISTS`.
- PostgreSQL would actually reject `DROP TYPE foo` if `foo` is used in a table somewhere; I did not add such checks in this PR.
- Only `AS ENUM (...)` is supported; the other PostgreSQL forms (composite, range, base, shell) are not.
- The similar `CREATE DOMAIN` statement, which allows aliasing existing types with some extra constraints is not supported.

### DIFF
--- a/lib/dialect.ml
+++ b/lib/dialect.ml
@@ -22,6 +22,7 @@ type feature =
   | DefaultExpr
   | Ttl
   | AlterColumn
+  | UserDefinedType
 [@@deriving show { with_path = false }]
 
 let show_feature x = 
@@ -45,6 +46,7 @@ let feature_to_string = function
   | DefaultExpr -> "default_expr"
   | Ttl -> "ttl"
   | AlterColumn -> "alter_column"
+  | UserDefinedType -> "user_defined_type"
 
 let feature_of_string s =
   match String.lowercase_ascii s with
@@ -63,6 +65,7 @@ let feature_of_string s =
   | "default_expr" -> DefaultExpr
   | "ttl" -> Ttl
   | "alter_column" -> AlterColumn
+  | "user_defined_type" -> UserDefinedType
   | _ -> failwith (Printf.sprintf "Unknown feature: %s" s)
 
 type support_state = {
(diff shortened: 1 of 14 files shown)


jongleb:
Rebased this on master, Made `TYPE` a real keyword instead of 

```
| DROP kw=IDENT if_exists? name=IDENT
     { if String.lowercase_ascii kw <> "type" then failwith ("expected TYPE after DROP, got " ^ kw);
       DropType name }
```
  
It is an unreserved keyword. https://github.com/postgres/postgres/blob/REL_18_0/src/backend/parser/gram.y 

Performing the replacement in the parser turned out it's noise, but I decided it was the better approach.

And renamed types.ml to user_types.ml to avoid confusion with the type inference

jongleb:
Thank you ! I merged

Niols:
Nice, thank you :-)

PR #284, opened 2026-06-11
Implement `JOIN` and subquery column elimination in dynamic select

jongleb:
### Description

This PR introduces `LEFT JOIN` elimination for dynamic selects. It also works for subqueries as sources, it means that any columns not used in the final select will be removed automatically."

For the user, there are two cases when it is convenient for the user to use this.

#### For example

Given this ddl

```sql
CREATE TABLE users (
  id INT PRIMARY KEY,
  org_id INT NOT NULL,
  name TEXT NOT NULL,
  email TEXT NOT NULL,
  created_at TIMESTAMP NOT NULL,
  deleted BOOLEAN NOT NULL
);
CREATE TABLE profiles (
  user_id INT PRIMARY KEY,
  bio TEXT,
  avatar_url TEXT,
  location TEXT,
  website TEXT
);
CREATE TABLE billing (
  user_id INT PRIMARY KEY,
  plan TEXT NOT NULL,
  paid_until DATETIME,
  balance INT NOT NULL
);
```

##### Before:

```sql
-- @user_brief
SELECT u.name, u.email
FROM users u
WHERE u.org_id = @org AND u.deleted = FALSE;

-- @user_card
SELECT u.name, p.bio, p.avatar_url
FROM users u
LEFT JOIN profiles p ON p.user_id = u.id
WHERE u.org_id = @org AND u.deleted = FALSE;

-- @user_admin
SELECT u.id, u.name, u.email, u.created_at,
       p.bio, p.avatar_url, p.location, p.website,
       b.plan, b.paid_until, b.balance
FROM users u
LEFT JOIN profiles p ON p.user_id = u.id
LEFT JOIN billing  b ON b.user_id = u.id
WHERE u.org_id = @org AND u.deleted = FALSE;
```

We could write three separate queries.Not only does the old approach force us to duplicate fields and parts of the query, but it also requires repeating the exact same `WHERE` conditions every time.

##### Since this PR

```sql
-- [sqlgg] dynamic_select=true
-- @user_info
SELECT u.id, u.name, u.email, u.created_at,
       p.bio, p.avatar_url, p.location, p.website,
       b.plan, b.paid_until, b.balance
FROM users u
LEFT JOIN profiles p ON p.user_id = u.id
LEFT JOIN billing  b ON b.user_id = u.id
WHERE u.org_id = @org AND u.deleted = FALSE;
```

And then it could be used this way:

```ocaml
let open Db.User_info_col in

(* 1. single column  *)
List.select db name ~org (fun n -> ...)
(* SELECT u.name FROM users u
   WHERE u.org_id = ? AND u.deleted = FALSE *)

(* 2. a pair of columns *)
List.select db (let+ n = name and+ b = bio in (n, b)) ~org (fun (n, b) -> ...)
(* SELECT u.name, p.bio
   FROM users u LEFT JOIN profiles p ON p.user_id = u.id
   WHERE u.org_id = ? AND u.deleted = FALSE *)

(* 3. any number of columns across tables and joins are pulled in
   automatically*)
List.select db
  (let+ n = name and+ b = bio and+ pl = plan and+ bal = balance in (n, b, pl, bal))
  ~org (fun (n, b, pl, bal) -> ...)
(* SELECT u.name, p.bio, b.plan, b.balance
   FROM users u LEFT JOIN profiles p ON p.user_id = u.id LEFT JOIN billing  b ON b.user_id = u.id
   WHERE u.org_id = ? AND u.deleted = FALSE *)
```

##### What kind of joins can be eliminated

Not every join can be eliminated. We can safely remove `LEFT JOIN` only if right table is joined by unique key and its columns are not used anywhere else in query. In this case join doesn't affect row count or filtering. Extra `AND` conditions on top of key match are fine, so `ON a.id = b.id AND b.type = 'x'` is still droppable. But if `ON` clause doesn't guarantee unique match (for example `ON a.id = b.id OR b.type = 'x'`, inequality or just part of composite key), we must keep this join. Otherwise it can multiply rows and change final result.

#### Subqueries

Before this PR the pick only affected the outer SELECT. The subquery always ran as written: all columns, all joins.

```sql
-- [sqlgg] dynamic_select=true
-- @user_info
SELECT * FROM (
  SELECT u.name, u.email, p.bio, b.plan
  FROM users u
  LEFT JOIN profiles p ON p.user_id = u.id
  LEFT JOIN billing b ON b.user_id = u.id
  WHERE u.org_id = @org AND u.deleted = FALSE
) AS sub;
```

Before: pick name, pay for everything anyway:

The resulting query
```sql
SELECT sub.name
FROM (SELECT u.name, u.email, p.bio, b.plan 
      FROM users u
      LEFT JOIN profiles p ON p.user_id = u.id     
      LEFT JOIN billing  b ON b.user_id = u.id    
      WHERE u.org_id = ? AND u.deleted = FALSE) AS sub
```

After the subquery narrows together with the pick

```ocaml

(* pick: name *)
List.select db name ~org (fun n -> ...)
(* SELECT * FROM (SELECT u.name
                  FROM users u
                  WHERE u.org_id = ? AND u.deleted = FALSE) AS sub *)

(* pick: name + bio *)
List.select db (let+ n = name and+ b = bio in (n, b)) ~org (fun (n, b) -> ...)
(* SELECT * FROM (SELECT u.name, p.bio
                  FROM users u LEFT JOIN profiles p ON p.user_id = u.id
                  WHERE u.org_id = ? AND u.deleted = FALSE) AS sub *)

(* pick: everything *)
List.select db
  (let+ n = name and+ e = email and+ b = bio and+ pl = plan in (n, e, b, pl))
  ~org (fun (n, e, b, pl) -> ...)
(* SELECT * FROM (SELECT u.name, u.email, p.bio, b.plan
                  FROM users u
                  LEFT JOIN profiles p ON p.user_id = u.id
                  LEFT JOIN billing b ON b.user_id = u.id
                  WHERE u.org_id = ? AND u.deleted = FALSE) AS sub *)
```

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -4,7 +4,7 @@ open Printf
 open ExtLib
 open Prelude
 
-type pos = (int * int) [@@deriving show]
+type pos = int * int [@@deriving show]
 
 type 'a located  = { value : 'a; pos : pos } [@@deriving show, make]
 type 'a collated = { collated: 'a; collation: string located option } [@@deriving show, make]
@@ -328,6 +328,17 @@ end
 type attr = {name : string; domain : Type.t; extra : Constraints.t; meta: Meta.t; }
   [@@deriving show {with_path=false}]
 
+let unique_keys schema =
+  let keys_of a =
+    Constraints.fold (fun c acc ->
+      match c with
+      | Constraint.PrimaryKey | Unique -> Constraint.StringSet.singleton a.name :: acc
+      | Composite (CompositePrimary s | CompositeUnique s) -> s :: acc
+      | NotNull | Null | Autoincrement | OnConflict _ | WithDefault -> acc)
+      a.extra []
+  in
+  List.concat_map keys_of schema |> List.sort_uniq Constraint.StringSet.compare
+
 let make_attribute name kind extra ~meta =
   if Constraints.mem Null extra && Constraints.mem NotNull extra then fail "Column %s can be either NULL or NOT NULL, but not both" name;
   let domain = Type.{ t = Option.default Int kind; nullability = if List.exists (fun cstrt -> Constraints.mem cstrt extra) [NotNull; PrimaryKey]
@@ -350,6 +361,8 @@ struct
       type 'a t = { attr: attr; sources: 'a list } [@@deriving show]
 
       let by_name name sattr = sattr.attr.name = name
+
+      let map_attr f sattr = { sattr with attr = f sattr.attr }
     end
 
     type 'a t = 'a Attr.t list
@@ -463,7 +476,7 @@ struct
     List.combine t1 t2
     |> List.mapi begin fun i (a1,a2) ->
       match Type.supertype a1.attr.domain a2.attr.domain with
-      | Some t -> { a1 with attr = { a1.attr with domain=t } }
+      | Some t -> Attr.map_attr (fun attr -> { attr with domain = t }) a1
       | None -> raise (Error (List.map (fun i -> i.attr) t1, sprintf "Attributes do not match : %s of type %s and %s of type %s"
         (show_name i a1.attr) (Type.show a1.attr.domain)
         (show_name i a2.attr) (Type.show a2.attr.domain)))
@@ -505,6 +518,9 @@ let show_table_name { db; tn } = match db with Some db -> sprintf "%s.%s" db tn 
 let make_table_name ?db tn = { db; tn }
 type schema = Schema.t [@@deriving show]
 type table = table_name * schema [@@deriving show]
+
+type join_source = { table : table_name; alias : table_name option } [@@deriving show]
+let join_source_name { table; alias } = Option.default table alias
 
 let print_table out (name,schema) =
   IO.write_line out (show_table_name name);
@@ -581,6 +597,7 @@ and var =
 | ChoiceIn of { param: param_id; kind : in_or_not_in; vars: var list }
 | Choice of param_id * ctor list
 | DynamicSelect of param_id * ctor list
+| DynamicSelectJoin of { pid : param_id; pos : pos; source : join_source }
 | TupleList of param_id * tuple_list_kind
 (* It differs from Choice that in this case we should generate sql "TRUE", it doesn't seem reusable *)
 | OptionActionChoice of param_id * var list * (pos * pos) * option_actions_kind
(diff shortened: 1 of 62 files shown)


lib/sql.ml
```diff
@@ -701,6 +750,40 @@ let source_fun_kind_to_infer = function
...
+    @ option_list else_
+
+let map_sub_exprs f = function
+  | Value _ | Param _ | Inparam _ | Column _ | Of_values _ | SelectExpr _ as e -> e
+  | Choices (n, l) -> Choices (n, List.map (fun (n, e) -> n, Option.map f e) l)
+  | InChoice (n, k, e) -> InChoice (n, k, f e)
+  | OptionActions ({ choice; _ } as o) -> OptionActions { o with choice = f choice }
+  | Fun ({ kind = Agg (With_order ({ order; _ } as wo)); parameters; _ } as fn) ->
+    Fun { fn with
+          kind = Agg (With_order { wo with order = List.map (fun (e, dir) -> f e, dir) order });
```
jongleb, line 785: I was writing a tree walkthrough again and realized this is time to move it to a separate place. 
Besides, adding a new expression forces us to go through every places adding a new match

https://gitlab.inria.fr/fpottier/visitors
and by the way  it’s even easier to get all these maps and traverses by adding this ppx

lib/sql.ml
```diff
@@ -580,6 +597,38 @@ and tuple_list_kind =
...
+  | ChoiceIn t -> ChoiceIn { t with vars = f t.vars }
+  | OptionActionChoice (p, vars, pos, kind) -> OptionActionChoice (p, f vars, pos, kind)
+  | SharedVarsGroup (vars, id) -> SharedVarsGroup (f vars, id)
+  | Choice (p, ctors) -> Choice (p, List.map map_ctor ctors)
+  | DynamicSelect (p, ctors) -> DynamicSelect (p, List.map map_ctor ctors)
+
+let var_pos = function
+  | Single (p, _) | SingleIn (p, _) -> fst p.id.pos
+  | Choice (id, _) | DynamicSelect (id, _) | TupleList (id, _)
+  | OptionActionChoice (id, _, _, _) -> fst id.pos
```
jongleb, line 639: https://github.com/ygrek/sqlgg/pull/284#discussion_r3403212259
taken out separately for the same reason

PR #288, opened 2026-06-29
Also attach column `[sqlgg]` metadata to quoted identifiers

Niols:
The `-- [sqlgg] key=value` annotation placed before a column was lost whenever the column name was a quoted identifier (`"id"`, `'id'`, `$tag$..`).

The statement splitter (`ruleStatement` in `sql_lexer.mll`) emits quoted identifiers as a single `` `Token ``, whereas unquoted names arrive as a stream of `` `Char ``. `extract_statement'` only flushed the pending column metadata in the `` `Char `` branch, so for a quoted column the metadata was recorded at the offset of the *next* `` `Char `` (e.g. the column type) and never matched the column's `$startpos` in the parser — silently dropping the annotation.

Flush the pending metadata in the `` `Token `` branch too (factored into `flush_internal_props`). This notably fixes `module=` type mappings for PostgreSQL schemas, which conventionally double-quote every identifier.

### DIFF
--- a/src/main.ml
+++ b/src/main.ml
@@ -171,21 +171,24 @@ let extract_statement' tokens =
     let answer () = Buffer.contents b, !props in
 
     let internal_props = ref Props.empty in
-    
+
+    let flush_internal_props () =
+      if List.length !internal_props > 0 then (
+        Parser_state.Stmt_metadata.add (Buffer.length b) !internal_props;
+        internal_props := Props.empty;
+      )
+    in
+
     let rec loop smth =
       match Enum.get tokens with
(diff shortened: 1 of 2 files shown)


jongleb:
Thank you!
just nitpick:

`!internal_props <> Props.empty` instead of `List.length !internal_props > 0` and I am merging

Niols:
Replaced `List.length !internal_props > 0` by `not (List.is_empty !internal_props)`. If you prefer `<> Props.empty`, I can also do that, but I try to avoid polymorphic comparison as much as possible.

Niols:
Also rebased on most recent master.

PR #291, opened 2026-07-04
Add support for PostgreSQL's trigram operators

Niols:
Adds parsing and typing for [PostgreSQL's pg_trgm operators](https://www.postgresql.org/docs/current/pgtrgm.html):

- Similarity operators `<%`, `%>`, `<<%`, `%>>` → new `TEXT_CMP_OP` (comparison, `Comp_text_cmp`).
- Distance operators `<->`, `<<->`, `<->>`, `<<<->`, `<->>>` → new `TEXT_DIST_OP` (returns float).
- The `similarity`, `word_similarity` and `strict_word_similarity` functions (`text × text → float`).

Note: I deliberately left out the `%` trigram similarity operator. The difficulty is that `%` is overloaded across dialects: it's the integer modulo `INT -> INT -> INT` in both MySQL and PostgreSQL, but in PostgreSQL it *also* doubles as the trigram similarity operator `TEXT -> TEXT -> BOOL`. I don't think there is currently a good way to treat such a case, but we would need to have:

- Some form of union type for operators within the same dialect (eg. PostgreSQL's `%s` should have type `(INT -> INT -> INT) U (TEXT -> TEXT -> BOOL)`).
- Have different types for the same operators in different dialects, or even a way to make operators available only for one dialect.

I left it as a FIXME in the lexer, and I'm willing to work on this some time if you think this is worth pursuing.

### DIFF
--- a/lib/sql.ml
+++ b/lib/sql.ml
@@ -656,7 +656,7 @@ type col_name = {
   tname : table_name option;
 } [@@deriving show]
 type logical_op = And | Or | Xor [@@deriving show]
-type comparison_op = Comp_equal | Comp_num_cmp | Comp_num_eq | Not_distinct_op | Is_null | Is_not_null [@@deriving eq, show]
+type comparison_op = Comp_equal | Comp_num_cmp | Comp_text_cmp | Comp_num_eq | Not_distinct_op | Is_null | Is_not_null [@@deriving eq, show]
 type null_handling_fn_kind = Coalesce of Type.tyvar * Type.tyvar | Null_if | If_null [@@deriving show]
 type source_alias = { table_name : table_name; column_aliases : schema option } [@@deriving show]
 type select_row_locking_kind = For_update | For_share [@@deriving show]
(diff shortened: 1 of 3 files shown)


lib/sql.ml
```diff
@@ -656,7 +656,7 @@ type col_name = {
   tname : table_name option;
 } [@@deriving show]
 type logical_op = And | Or | Xor [@@deriving show]
-type comparison_op = Comp_equal | Comp_num_cmp | Comp_num_eq | Not_distinct_op | Is_null | Is_not_null [@@deriving eq, show]
+type comparison_op = Comp_equal | Comp_num_cmp | Comp_text_cmp | Comp_num_eq | Not_distinct_op | Is_null | Is_not_null [@@deriving eq, show]
```
ygrek, line 659: i was going to say i would expect some typing changes involving new ctor, but then i see there is nothing like this for comp_num_cmp either. Not for this PR but it means this is a missed opportunity to have better inference @jongleb ?

ygrek (approved):

ygrek:
as for % ambiguity i think we already have some ambiguity handling but it is case by case handling, also it will be the first one to depend on dialect i think yes

Niols:
If this was approved, should we merge it, and move the discussion on operators inference to an issue?

lib/sql.ml
jongleb: Sure, got it!

PR #292, opened 2026-07-04
Fix numbering of PostgreSQL params across reusable queries

Niols:
When a reusable query (`-- @foo | include: reuse`) is spliced into an outer query as a CTE via `WITH x AS &ref`, the reused query's parameters and the outer query's parameters end up in a single generated statement, bound positionally by `set_params`.

With `-params postgresql`, the positional counter was reset to `0` for the spliced fragment and left unchanged when the outer query resumed, so both fragments numbered their `$N` placeholders starting from `$1`. The outer parameters therefore collided with the reused ones. Example:

``` sql
-- @get_persons | include: reuse
SELECT * FROM person WHERE name LIKE @name;

-- @get_users 
WITH person AS &get_persons
SELECT * FROM "user" JOIN person ON person.user_id = "user".id
WHERE "user".id > @min_id;
```

On such an input, `-params postgresql` generated:

``` sql
WITH person AS (... WHERE name LIKE $1)
SELECT ... WHERE "user".id > $1
```

Note the `$1` used twice.

(This was actually caught by PostgreSQL because `set_params` still supplies a value for every position, so I got “bind message supplies N parameters, but prepared statement requires M”.)

I checked this only affects the `postgresql` parameter style; others use names (`named` and `oracle`) or use positional `?` (`unnamed`).

A first commit adds tests exhibiting this behaviour; a second threads the positional counter through the shared-query splice in `substitute_vars`.

### DIFF
--- a/src/gen.ml
+++ b/src/gen.ml
@@ -90,7 +90,7 @@ and sql_dynamic_ctor = {
 let substitute_vars s vars subst_param =
   let rec loop s acc i parami vars =
     match vars with
-    | [] -> acc, i
+    | [] -> acc, i, parami
     | Sql.Single (param, _) :: tl ->
       let (i1,i2) = param.id.pos in
       assert (i2 > i1);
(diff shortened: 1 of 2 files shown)


Niols:
Anything I can do to help this PR move forward?

jongleb:
@Niols Thank you!

PR #295, opened 2026-07-14
Support shared `Choices` expr and make Dynamic select compose with `CTE` and `IN` exprs

jongleb:
### Description

This PR adds the support of shared Choices and makes dynamic select abosuletly composable with the rest of the exprs

#### Same choice can now be used several times

```sql
SELECT * FROM t
WHERE a = @x { A { 1 } | B { 2 } }
  AND b = @x { A { 10 } | B { 20 } };
```

This query was failed before since `@x` Choices expr was presented several times. now it's fixed.

####  Param names inside `Choices` branches don't have to match now

```sql
WITH cte AS (
 SELECT id FROM t 
 WHERE @f { ByStatus { status = @s } | All { TRUE } }
)
SELECT id FROM cte WHERE @f { ByStatus { id > @min } | All { TRUE } };
```

`@s` and `@min` are the same parameter here 

####  Dynamic select now composes with everything rest

```sql
-- @fragment_mixed | include: reuse
SELECT id, name, status FROM t
WHERE id IN @ids AND { status = @st }? AND @c { All { TRUE } | ByS { status = @s } };

-- [sqlgg] dynamic_select=true
-- @reuse_mixed
WITH f AS &fragment_mixed
SELECT id, name
FROM f
WHERE (id, name) IN @pairs
ORDER BY @sort { I { id } | N { name } };
```

Now: IN `@ids`, CTEs and reusable CTEs all work

query produced at runtime (checked by a [print test](https://github.com/jongleb/sqlgg/blob/branches-composable/test/cram/test_dynamic_select_matrix/run.t#L171)):

```sql
WITH f AS (SELECT id, name, status FROM t
WHERE id IN (111, 112) AND  (  status = 113  )  AND  ( status = 114 ) )
SELECT name
FROM f
WHERE (id, name) IN ((333, 'a'))
ORDER BY  ( name )
```

### DIFF
--- a/lib/syntax.ml
+++ b/lib/syntax.ml
@@ -277,12 +277,13 @@ let dynamic_allowed env =
   | Top_level | From_passthrough -> true
   | Subquery -> false
 
+let dynamic_col_param_name = "col"
+
 let make_dynamic_select ~env columns =
   if not (dynamic_allowed env) then
     columns
   else
     let module S = Set.Make(String) in
-    let dynamic_col_param_name = "col" in
     let unique_name used base =
       if not (S.mem base used) then
         base
@@ -1402,7 +1403,7 @@ and eval_select_full env { select_complete; cte } =
 and eval_cte { cte_items; is_recursive } = 
   let open Schema.Source in
   List.fold_left begin fun (acc_ctes, acc_vars) cte ->
-    let env = { empty_env with ctes = acc_ctes } in
+    let env = { empty_env with ctes = acc_ctes; scope = Subquery } in
     let tbl_name = make_table_name cte.cte_name in
     let a1 = List.map (fun attr -> Attr.{ sources = []; attr }) in
     let s1, p1, _kind =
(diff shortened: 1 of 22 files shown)


lib/syntax.ml
```diff
@@ -1848,50 +1849,131 @@ let rec eval (stmt:Sql.stmt) =
...
+
+  let node t name =
+    match Hashtbl.find_opt t name with
+    | Some n -> n
+    | None -> let n = { name; state = Root None } in Hashtbl.add t name n; n
+
+  let rec root n =
+    match n.state with
+    | Root typ -> n, typ
+    | Link p -> let (r, _) as res = root p in if r != p then n.state <- Link r; res
```
ygrek, line 1902: physical comparison on purpose? must comment

lib/syntax.ml
```diff
@@ -1848,50 +1849,131 @@ let rec eval (stmt:Sql.stmt) =
...
+    | None -> fail "incompatible types for parameter %S : %s and %s" name (Type.show t1) (Type.show t2)
+
+  let note t name typ =
+    let (r, existing) = root (node t name) in
+    r.state <- Root (Some (Option.map_default (unify name typ) typ existing))
+
+  let alias t n1 n2 =
+    let (r1, _) = root (node t n1) in
+    let (r2, t2) = root (node t n2) in
+    if r1 != r2 then begin
```
ygrek, line 1918: same

src/gen_caml.ml
```diff
@@ -155,22 +150,13 @@ type scoped_field = {
...
-  | { t = One_or_all; _ }
-  | { t = UInt64; _ }
-  | { t = Any; _ } as t -> type_name t
+  let as_lang_type t =
+    type_name @@
+    match t with
+    | { t = Blob; nullability } | { t = StringLiteral _; nullability } -> { t = Text; nullability }
+    | { t = FloatingLiteral _; nullability } -> { t = Float; nullability }
+    | { t = Decimal _; nullability } -> { t = Decimal { precision = None; scale = None }; nullability }
+    | t -> t
```
ygrek, line 159: i think explicit enumeration was on purpose

lib/syntax.ml
jongleb: Yes , like here
https://gitlab.inria.fr/fpottier/unionfind/-/blob/master/src/UnionFindBasic.ml#L64-72

lib/syntax.ml
jongleb: Got it

src/gen_caml.ml
jongleb: Okay, I'll return, got it

PR #308, opened 2026-08-07
Get rid of dependency on `mybuild`

Niols:
The `mybuild` dependency came only to provide the version numbers. I see two ways to replace it:

- Rely on `dune subst` and simply have `%%VERSION%%` in the code. This is the simplest but it will actually print the string `%%VERSION%%` during development as `dune subst` only runs during actual installation.

- Rely on something like [`dune-build-info`](https://dune.readthedocs.io/en/stable/dune-libs.html#dune-build-info-library) which can do it but would introduce another dependency.

This PR implements the former. While I'm at it, I also add the explicit dependency in `unix` to silence the corresponding >5.0 warning.

### DIFF
--- a/impl/ocaml/dune
+++ b/impl/ocaml/dune
@@ -2,4 +2,4 @@
  (wrapped false)
  (name sqlgg_traits)
  (public_name sqlgg.traits)
- (libraries integers sqlgg_json_path))
+ (libraries integers sqlgg_json_path unix))
--- a/sqlgg.opam
+++ b/sqlgg.opam
@@ -16,7 +16,6 @@ depends: [
   "dune" {>= "2.7"}
   "integers" {>= "0.4.0"}
   "menhir" {>= "20180523"}
-  "mybuild" {> "3"}
   "ppx_deriving" {>= "4.3"}
   "ppx_enumerate" {>= "v0.15.0"}
   "ppx_deriving_variant_string" {>= "1.0.1"}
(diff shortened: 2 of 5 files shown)


ygrek (changes requested):
need to update opam file as well

ygrek:
how does it detect version? what it will look like now?

Niols:
It will first look at metadata in the `dune-project` or opam files and use those. If they aren't set it asks Git with:
```
git describe --always --dirty --abbrev=7
```
which produces something of the form
```
<version>-<commits-since-version>-<hash>[-dirty]
```
[[Source](https://dune.readthedocs.io/en/stable/advanced/package-version.html); [Source](https://dune.readthedocs.io/en/stable/dune-libs.html#build-info)]


For instance, as of this commit:
```
$ git describe --always --dirty --abbrev=7
20231201-385-g88c420e

$ _build/install/default/bin/sqlgg 
SQL Guided (code) Generator ver. 20231201-385-g88c420e
Usage: cli.exe <options> <file.sql> [<file2.sql> ...]
Options are:
[...]
```
(It ignores the `20260721` because `git describe` only considers _annotated_ tags, but we could easily recreate the tag as an annotated one.)

Niols:
> need to update opam file as well

I'm not sure what you want me to change in the opam file? I already removed `mybuild` from there.

ygrek:
oh i am blind

PR #313, opened 2026-08-23
add -open <file>: load definitions without generating code for them

andreypopp:
-open parses the file at its position on the command line for its side
effects (schema, include:reuse queries) but emits no code, so shared
definitions can back several generated modules without duplication.
A file both -open'ed and given as an input is processed once, as a
regular input (files are identified by canonical path), which makes a
single command shape work for every file of a project - e.g. as a dune
dialect preprocessor, see test/cram/dune_dialect.t.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jongleb (approved):
Nice! Thanks!

PR #318, opened 2026-08-27
add support cached tables (ALTER TABLE ... CACHE / NOCACHE)

ProgramingIsTheFuture:
Support cached tables from tiDB: https://docs.pingcap.com/tidb/stable/cached-tables/

Examples:
```
ALTER TABLE foo CACHE;
ALTER TABLE foo NOCACHE;
```


jongleb (approved):
it is done using dialects that is LGTM

PR #329, opened 2026-09-15
Add support for PostgreSQL's CREATE/DROP EXTENSION

Niols:
Both statements are accepted with -dialect postgresql, so that migration files that install pg_trgm and friends can be fed to sqlgg unchanged.

sqlgg keeps no extension state, so there is nothing to track: the optional SCHEMA, VERSION and CASCADE clauses are parsed and discarded, an extension can be dropped without having been created, and the statements evaluate to Stmt.Other rather than getting a statement kind of their own.

They are gated behind a new `extension` dialect feature, PostgreSQL only, like the existing `user_defined_type` gate on CREATE TYPE.

`extension`, `schema` and `version` become keywords, so they are added to `ident` to keep them unreserved, the way `type` already is.

CREATE EXTENSION's trailing `[ WITH ] ...` clause is ambiguous with a following statement that starts with a WITH (a CTE), which LR(1) cannot resolve. Both statements therefore live in a new `top_statement` rule used only by `input`, where EOF bounds the clause. Extensions are database-global DDL that has no business in a routine body anyway.


jongleb:
Thank you! I added a follow-up to keep extension, schema, and version usable as identifiers in the parser and LSP. Merged

Niols:
Thank you!

